PR #44, opened 2025-04-23
Test Cleanup

fpham0701:
- Cleaned up the test cases (differentiating valid and invalid ones for interpreter)
- Added `while` and `if` test case for interpreter
- Added snapshot testing for interpreter
- Updated snapshots
- Minor fix to parsing (if no step value is indicated then send a warning) and change of Binary Operations output logic in typecheck


ekiwi (approved):
:shipit:

PR #48, opened 2025-05-19
Scheduling Improvements

Nikil-Shyamsunder:
- Identity function verilog


tests/identities/dual_identity_d1.prot
```diff
@@ -0,0 +1,42 @@
...
+}
+
+fn one<dut: DualIdentity>(in a: u32, out s1: u32, out s2 : u32) {
+    dut.a := 1;
+    dut.b := X;
+    fork();
+
+    step();
+
+    if (dut.a == a) {
```
ekiwi, line 15: can you add a comment here to explain why this `if` construct here is relevant for your test?
on its own, this protocol is not self-explanatory.

tests/identities/dual_identity_d1.prot
```diff
@@ -0,0 +1,42 @@
...
+
+    step();
+
+    assert_eq(dut.s1, s1);
+    assert_eq(dut.s2, s2);
+}
+
+
+fn two<dut: DualIdentity>(in b: u32, out s1: u32, out s2 : u32) {
+    if (dut.b == b) {
```
ekiwi, line 30: can you add a comment here to explain why this `if` construct here is relevant for your test?
on its own, this protocol is not self-explanatory.

tests/identities/identity_d2_multiple_assign.prot
```diff
@@ -0,0 +1,13 @@
+struct Identity {
+  in a: u32,
+  out s: u32,
+}
+
+fn identity<DUT: Identity>(in a: u32, out s: u32) {
+  DUT.a := a;
+  fork();
+  step();
+  DUT.a := a;
```
ekiwi, line 10: This is always going to end in an error if you execute more than one instance of this protocol, right?

tests/snapshots/protocols__interpreter__tests__mult_execution.snap
```diff
@@ -2,10 +2,4 @@
...
 expression: content
 ---
-Assertion Failed
-
-error: The two expressions did not evaluate to the same value
-   ┌─ tests/mul.prot:19:13
-   │
-19 │   assert_eq(s, dut.s);
-   │             ^^^^^^^^ LHS Value: ValueOwned(00000000000000000000000000110000), RHS Value: ValueOwned(00000000000000000000000000000000)
+Assertion Passed
```
ekiwi, line 5: Why does this pass now?

tests/snapshots/protocols__interpreter__tests__add_ok.snap
```diff
@@ -2,10 +2,4 @@
...
 expression: content
 ---
-Assertion Failed
-
-error: The two expressions did not evaluate to the same value
-   ┌─ tests/add_struct.prot:17:13
-   │
-17 │   assert_eq(s, DUT.s);
-   │             ^^^^^^^^ LHS Value: ValueOwned(00000000000000000000000000001110), RHS Value: ValueOwned(00000000000000000000000000000000)
+Assertion Passed
```
ekiwi, line 5: Why does this pass now?

tests/snapshots/protocols__interpreter__tests__mult_execution.snap
Nikil-Shyamsunder: These tests no longer exist now that all tests have been moved out of the interpreter, so their snapshots are meaningless. I have now deleted all of the old interpreter test snapshots

Nikil-Shyamsunder: The expected outcomes occur correctly in the corresponding scheduler test

ekiwi: great!

ekiwi (changes requested):
Sorry, looks like you are still working on the comments

tests/identities/identity_d2_multiple_assign.prot
Nikil-Shyamsunder: Yes -- assuming that the two todos have their argument `a` set to different values. In `test_scheduler_identity_d2_multiple_assign` I test:
1. just one thread (passes)
2. two threads assigning different `NewValue`s (fails)
3. two threads assigning same `NewValue` (passes)

Issue #55, opened 2025-05-21
implicit forks for transactions


Nikil-Shyamsunder:
check with a flag if a thread has already forked, if it has it cannot fork again (dynamic check), if it doesn't fork upon completion perform an extra fork

Issue #35, opened 2025-03-20
Change the type_args field in Transaction to be type_param with type symbolid


Nikil-Shyamsunder:
there is no need for the type_args field to be a `Vec` since transactions are restricted to a single DUT type parameter

PR #56, opened 2025-05-21
Code Cleanup + Forking 

Nikil-Shyamsunder:
Adressing #57:
- Added Clippy to CI with the rust version pinned
- resolved all clippy linting errors
- cleaned up a lot of code in the scheduler and interpreter. this included getting rid of public attributes in some of the structs and replacing them with proper mutator functions, creating some type aliases and helper structs, adding helper functions, moving certain functionality from the scheduler into the interpreter where it makes more sense, and generally improving code readability across these two files. 

Addressing #35:
- changed the `type_args` vector to be a single `type_param` of type `Option<SymbolId>`. The `type_param` starts off as `None` but after parsing is guaranteed to be `Some` since a type argument is required in the grammar.

Addressing #55:
- each `Thread` has a `has_forked` flag. At runtime, if a fork statement is reached this flag becomes true. if another fork statement is reached, an error is thrown and execution of the thread exits. Note that this is *independent of whether another todo to fork actually exists*. Is that the correct semantics?
- added a check such that if a thread has completed successfully--that is, if forks are enabled, and this thread has None for its next_step (execution is completed), and the thread didn't error out, fork. Is that the correct semantics?


src/interpreter.rs
```diff
@@ -110,17 +110,15 @@ impl<'a> Evaluator<'a> {
...
-                    // Only process outputs that have a name
-                    ctx.get_symbol_name((o).expr).map(|name| (o, name))
-                })
-                .find(|(_, name)| *name == symbol_name)
+                .map(|o| ctx[o.name].to_string())
+                .enumerate()
+                .find(|(_, out_name)| out_name == symbol_name)
             {
-                output_mapping.insert(*symbol_id, *output_ref.0);
+                output_mapping.insert(*symbol_id, sys.outputs[idx]);
```
ekiwi: @Nikil-Shyamsunder There might be a better way to do this.

The old code was decent, with the one problem that `get_symbol_name` does not make sense for outputs, since outputs are rarely symbols but more complex expressions instead. Instead, you want to look at the `o.name` as you are doing in the new structure.

Here is a sketch for solving this more elegantly:
```.rs
let output = sys.outputs.iter().find(|o| ctx[o.name] == symbol_name).expect("failed to find matching output");
output_mapping.insert(*symbol_id, output);
```

ekiwi (approved):

Issue #52, opened 2025-05-21
can we allow if statements without an else block (infer an else with an empty block)


Nikil-Shyamsunder:

ekiwi:
That would be a useful thing to add.

PR #60, opened 2025-05-22
Allow Empty Blocks

Nikil-Shyamsunder:
We are trying to address #52 to allow if statements without a corresponding else block. the solution is to make that just syntactic sugar for an if followed by an empty block.

### Change parser to allow if-only statements and parse into de-sugared version.
The grammar currently allows for an if-only statement, but the parser expects an else. Here's the fix:
```rust
-        let else_rule = self.expect_rule(inner_rules.next(), &pair, "Expected else block")?;
-        let inner_else = else_rule.into_inner();
-        let else_block = self.parse_stmt_block(inner_else)?;
+
+        // Parse the optional else block
+        let else_block = inner_rules
+            .next()
+            .map(|else_rule| self.parse_stmt_block(else_rule.into_inner()))
+            .transpose()?
+            .unwrap_or_else(|| self.tr.s(Stmt::Block(vec![])));
         Ok(Stmt::IfElse(expr_id, if_block, else_block))
```

Testing: 
`test_simple_if_without_else_transaction`:This is just a basic serialization transaction, where we have an if-only input, and expect a snapshot output that has the identical if followed by an empty else block. 
**Input:**
```
fn if_without_else<dut: DualIdentity>(in a: u64, in b: u64) {
    if (dut.b == b) {
        dut.a := a;
    }
}
```
**Output:**
```
fn if_without_else<dut: DualIdentity>(in a: u64, in b: u64) {
  if dut.b == b {
    dut.a := a;
  } else {
  }
}
```

### Allowing for empty blocks
Step 1 is allowing for empty blocks in the first place; they are allowed in the grammar, but previously, an empty block was not handled properly by the scheduler due to an error in how the next statement mapping was generated and used by the interpreter. This is fixed in this PR.

Testing: The functionality is tests in two separate ways. First `test_scheduler_dual_identity` has a test case in which we have an if-only block which failed prior to this change. I also created a very simple empty protocol with a corresponding combinational logic identity function circuit, which should always pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 Cargo.lock
 .idea/
 test_run/
-/tests/snapshots
 *.btor
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -329,7 +329,7 @@ impl<'a> Evaluator<'a> {
             }
             Stmt::Block(stmt_ids) => {
                 if stmt_ids.is_empty() {
-                    return Ok(self.next_stmt_map[stmt_id]);
+                    Ok(self.next_stmt_map[stmt_id])
                 } else {
                     Ok(Some(stmt_ids[0]))
                 }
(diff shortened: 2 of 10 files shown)


ekiwi (approved):
Looks good

PR #67, opened 2025-05-26
Bitvec Width Type Inferencing

Nikil-Shyamsunder:
# bitvec width type inference
**First, regarding Constants in the IR**:  All constants are still stored as `BitVecValue`s inside the transaction-level IR. The parser assumes all constants are initially **128 bits wide**, so all constants in a `.prot` file are treated as unsigned and fit into a 128-bit vector at first.

### type inference overview
The type inference recurses over each expression in the `Transaction` and generates a set of constraints over each subexpression. There are some **observations I used to come up with this system:**

1. All operations on expressions require equal width operands and return a constant of the same width (FIX)
2. All symbols do not need type inference; we know there width's beforehand

### constraint generation
- **Binary Operations** (`AND`, `EQUAL`):  
  Require both operands to have **equal widths**.  
  -> Emit: `Eq(tv_left, tv_right)`
`AND` also requires the result to have equal width to the operands
`EQUAL` always returns a one-bit result. 

- **Unary Operations** (`NOT`):  
  Operand and result must have the same width.  
  -> Emit: `Eq(tv_operand, tv_this_expr)`

- **Constants**:  
  Emit a **LowerBound constraint** on their width, based on the minimum number of bits needed to represent the value.

- **Symbols**:  
  Emit both `LowerBound` and `UpperBound` constraints equal to the declared width.  

- **Assignment Statements**:  
  The RHS is constrained to exactly match the width of the LHS symbol, by introducing both `LowerBound` and `UpperBound` on the RHS

- **Slice Expressions**:  
  Emit a lower bound on the input expression (`high + 1`), and constrain the output to the exact width of the slice (`high - low + 1`).

### solving constraints using union-find
1. **Union-Find Over Type Variables**:
   - Each `Eq(a, b)` constraint is processed to merge equivalence classes of type variables.
   - After unifying, all type variables that must have the same width will share the same root in the union-find forest.

3. **Bound Collection and Validation**:
   - For each equivalence class (i.e., root node in union-find), the solver:
     - Gathers all `LowerBound` constraints -> takes the maximum (infimum)
     - Gathers all `UpperBound` constraints -> takes the minimum (supremum)
   - If `infimum > supremum`, the class is unsatisfiable, and an error is reported.

4. **Final Width Assignment**:
   - Each `TypeVar` is mapped to its **infimum**? I think this makes the most sense?
   - This mapping is used to resolve the width of every `ExprId`.

### finalization after inference
After solving:
- After parsing, we run the solver on each transaction. Each constant in the transaction is iterated over and truncate  (via the slice `BitVecOp` its `BitVecValue`) to the inferred bit-width. assuming our type inferencer doesn't throw an error that a class is unsatisfiable, this is guaranteed to never lose any information. 
- At this point, the rest of the code doesn't have to be changed at all! And constants are all of the right size.

**other small changes**:
- some functions took `Vec<Transaction, SymbolTable>` as argument, others took `Vec<SymbolTable, Transaction>`, so I standardized it to the former
- added the width inferencing and type checking to the `parse_file` function to complete the entire language's frontend pipeline.

### DIFF
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -83,7 +83,7 @@ impl Transaction {
         self.stmt_loc.get(stmt_id).copied()
     }
 
-    fn expr_widths(&self, st: SymbolTable) -> HashMap<ExprId, usize> {
+    fn expr_widths(&self, st: SymbolTable) -> HashMap<ExprId, u32> {
         let mut type_ctx = TypeContext::new(st.clone(), self.clone());
         type_ctx.finalize()
     }
--- a/src/ir.rs
+++ b/src/ir.rs
@@ -5,7 +5,7 @@
 // author: Francis Pham <fdp25@cornell.edu>
 
 use crate::type_inference::TypeContext;
-use baa::BitVecValue;
+use baa::{BitVecOps, BitVecValue};
 use cranelift_entity::{entity_impl, PrimaryMap, SecondaryMap};
 use rustc_hash::FxHashMap;
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


src/ir.rs
```diff
@@ -240,7 +247,7 @@ pub enum UnaryOp {
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 pub enum Expr {
     // nullary
-    Const(BitVecValue),
+    Const(u128),
```
ekiwi, line 250: I would prefer if you could keep this a `BitVecValue`. You can always start with a 128-bit BitVecValue and then later narrow down the width after inference.

PR #70, opened 2025-08-15
Tests & Examples

Nikil-Shyamsunder:
- split up files into test directory and examples directory
- add a bunch more tests from the legacy repo
- add explicit bit vec width syntax (verilog-like)


ekiwi (approved):
This looks great!

I would like to get all the test code out of `scheduler.rs` at some point, but we can do that in a different PR.

PR #82, opened 2025-09-03
Port subset of unit tests to Turnt + Add Justfile

ngernest:
This PR ports a subset of the unit tests in `scheduler.rs` to Turnt snapshot tests (specifically the ones for adders, multipliers & identities). 

Some of the identity scheduler unit tests are a bit more complex (e.g. the ones which test implicit forks & slicing), so I'll convert those to Turnt in a future PR.

**Setup**:
- Run `uv tool install turnt` to install [Turnt](https://github.com/cucapra/turnt/tree/main) 
  - Note: this presumes you already have `uv` installed (if not, [follow these instructions](https://docs.astral.sh/uv/getting-started/installation/#pypi))
- To run the snapshot tests for `adder_d1`,`cd` into the [`protocols/tests`](./protocols/tests/) subdirectory, then do `turnt tests/adders/adder_d1/*.tx`

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,16 +1,11 @@
-default: lint build test
-
 # At the moment, we only run Turnt on the tests which have been configured for Turnt
 turnt-tests:
   cd protocols/tests && turnt adders/*/*.tx && turnt multipliers/mult_d2/*.tx && turnt identities/identity_d2/*.tx
 
-build:
-  cargo build
-
 # Runs all unit tests (via Cargo) & snapshot tests (via Turnt)
 test:
   cargo test 
-  just turnt-tests
+  @just turnt-tests
 
 # Applies Clippy's suggestions & formats the code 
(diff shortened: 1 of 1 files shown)


ngernest:
To simplify running both Cargo tests + Turnt tests at the same time, I've also added a [Justfile](https://github.com/casey/just)

After [installing `just`](https://just.systems/man/en/packages.html), you can run `just test` in the top-level directory to run both Cargo + Turnt tests. 

If you run `just`, by default it applies the linter (`cargo fmt` + Clippy), builds the code, and runs all unit + snapshot tests. 
(This allows `just` to act sort of like a [pre-commit hook](https://pre-commit.com).)

For a future PR, we may want to integrate `just` into CI so that we check if there are any failing snapshot tests (via Turnt) every time we make a PR.

PR #86, opened 2025-09-15
Pretty-Print Statements/Expressions in Scheduler Logs

ngernest:
Previously, our logs were in the following form:
```rust
[INFO  protocols::scheduler]   Evaluating statement: stmt7
```
Ideally, we'd be able to pretty-print `stmt7` so that we have better insight into what statements are being executed by the scheduler. 

This PR addresses this issue by adding pretty-printers for `Stmt` / `Expr`s based on the `StmtId` / `ExprId`. 

The logs now look like this: 
```rust
[INFO  protocols::scheduler]   Evaluating statement: `DUT.a := a`
```

For a more complete example of the logs, run:
```bash
$ cargo run --package protocols-interp -- --verilog protocols/tests/adders/adder_d1/add_d1.v \
        --protocol protocols/tests/adders/adder_d1/add_d1.prot \
        -t protocols/tests/adders/adder_d1/both_threads_pass.tx \
        --verbose
```

This produces the following log:
```rust
[INFO  protocols::scheduler] ==== Starting scheduling cycle 1, active threads: 1 ====
[INFO  protocols::scheduler] Running thread add from `step()` (stmt7)
[INFO  protocols::scheduler]   Evaluating statement: `{ DUT.a := a
    DUT.b := b
    step()
    DUT.a := X
    DUT.b := X
    assert_eq(s, DUT.s) }`
[INFO  protocols::scheduler]   Next statement: stmt1 `DUT.a := a`
[INFO  protocols::scheduler]   Evaluating statement: `DUT.a := a`
[INFO  protocols::scheduler]   Next statement: stmt2 `DUT.b := b`
[INFO  protocols::scheduler]   Evaluating statement: `DUT.b := b`
[INFO  protocols::scheduler]   Next statement: stmt3 `step()`
[INFO  protocols::scheduler]   `Step()` reached at stmt3, pausing.
...
```

PR #87, opened 2025-09-15
Update pretty-printer for scheduler logs to use existing serializer

ngernest:
Updates the pretty-printer for statements/expressions in scheduler logs (#86) to use the existing functions in `serializer.rs`.

I've also rewritten some functions in `serializer.rs` to be a bit more idiomatic (e.g. implement `Display` trait where applicable).


fpham0701 (approved):
Looks good to me! Thanks for tidying this up, it seems much better

PR #89, opened 2025-09-16
WIP: Port more scheduler tests to Turnt

ngernest:


ekiwi (changes requested):
Looks good. However, instead of moving things out of examples into tests (or copying), could you instead point turnt to the examples directory? It might also make sense to just do two different tasks in the CI with two different turnt files, one for the tests and one for the examples.

ngernest:
@ekiwi I updated Turnt so that it also points to the `examples` directory! I also updated the random number generator so that we now instantiate it with the seed `0` (per #92), this involved changing the RNG from `ThreadRng` to [`StdRng`](https://rust-random.github.io/rand/rand/rngs/struct.StdRng.html) since `ThreadRng` doesn't support seeding.

PR #95, opened 2025-09-19
WIP: Keep running monitor when there are steps left in the signal trace

ngernest:


ekiwi (changes requested):
@ngernest: This looks good. However, there seem to be one or more bugs that you need to fix before we merge this.

From looking at the waveform, the output of your monitor should be:
```
add(1, 2, 3)
add(4, 5, 9)
```

Currently the output `s` is always `0` in your trace. You also print out 4 instead of 2 transactions. The final transaction, which starts in step 2 (2s .. 3s) is never completed, so you should not include it with the completed transactions.

ekiwi:
> 1. To distinguish whether a transaction is complete or not, should I find a way to first determine the no. of steps for the protocol

No. You are executing the transaction in your "backward interpreter", right? However, the backwards execution for the last transaction should never finish, since there isn't another stop available when it calls `step`. That way you should be able to tell, that the transaction never completed. To put it the other way around: you know that a transaction is completed when you reach the last statement of that transaction and sucessfully execute it.



> 2\. `DUT.s` isn't assigned a value in

For now, I would suggest that you print out `???` if you do not have a value for a given argument.

PR #96, opened 2025-10-01
Add checks for `ForkBeforeStep` + `DoubleFork` errors (+ docs on well-formedness)

ngernest:
This PR makes the following changes:
- Adds docs for well-formedness checks (`well-formedness.md`)

- If we try to call `fork` without having previously called `step`, the interpreter throws an error:
```rust
error: Thread 0 (transaction 'add_fork_before_step') called `fork()` without previously calling `step()` first
   ┌─ adders/adder_d1/fork_before_step_error.prot:11:3
   │
11 │   fork();
   │   ^^^^^^^ Thread 0 (transaction 'add_fork_before_step') called `fork()` without previously calling `step()` first
```

- The error messages for `DoubleFork` errors now refer to the locations of the two forks to aid the user, e.g.
```rust
error: Thread 0 (transaction 'add_double_fork') attempted to fork more than once
   ┌─ adders/adder_d1/double_fork_error.prot:12:3
   │
12 │   fork();
   │   ^^^^^^^ first fork() called here
   ·
15 │   fork();
   │   ^^^^^^^ second fork() called here

```
- ^^ This required adding methods to `Diagnostic` to support referring to multiple statements within the same error message

- Made the output waveform file an _optional_ parameter in the interpreter CLI (this is so that we don't create a new `trace.fst` file for every Turnt test that is run)

**TODOs for a future PR**
- No `fork`s in program (per 9/30 meeting, a function must have exactly one `fork` somewhere in its body)
- Missing `step` at the end of a function
(I'm doing them in a separate PR since they involve modifying some existing Turnt tests in order for them to pass.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Options:
   -m, --module <MODULE_NAME>
           Name of the top-level module (if one exists)
   -f, --fst <WAVEFORM_FILE>
-          Name of the waveform file (.fst) in which to save results [default: trace.fst]
+          (Optional) Name of the waveform file (.fst) in which to save results
   -v, --verbose...
           Prints logs / debugging information to stdout
   -h, --help
--- a/interp/src/main.rs
+++ b/interp/src/main.rs
@@ -33,16 +33,16 @@ struct Cli {
     #[arg(short, long, value_name = "MODULE_NAME")]
     module: Option<String>,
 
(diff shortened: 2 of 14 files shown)


fpham0701 (approved):
Seems good to me!

Issue #97, opened 2025-10-01
Reset all values to `DontCare` at the end of a function


ngernest:
When one protocol finishes, we want to reset all its pins to `DontCare` to limit information flow from one protocol to another. This should be done in the last cycle, i.e. we throw an error if people are performing assignments in the last cycle.

ngernest:
Resolved in #103

PR #103, opened 2025-10-06
WIP: Treat unused variables as DontCares + reset all values to DontCare at the end of a function

ngernest:
Addresses issue #97


ekiwi (approved):
Let's get this in, but continue investigating why the `mul_no_reset` protocol stuff does not fail.

PR #110, opened 2025-10-22
WIP: ALU Turnt tests for interpreter + monitor

ngernest:
```bash
$ cd monitor/tests
$ cargo run --package protocols-monitor -- --color never -p alus/alu_d1_monitor.prot --wave alus/alu_d1.fst --instances alu_d1:ALU
```

### DIFF
--- a/monitor/src/designs.rs
+++ b/monitor/src/designs.rs
@@ -6,9 +6,8 @@
 #![allow(dead_code)]
 
 use log::info;
-use protocols::ir::{Field, Ident, SymbolId, SymbolTable, Transaction, Type};
 use protocols::{
-    ir::{Field, SymbolId, SymbolTable, Transaction, Type},
+    ir::{Field, Ident, SymbolId, SymbolTable, Transaction, Type},
     serialize::serialize_field,
 };
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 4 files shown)


monitor/src/interpreter.rs
```diff
@@ -63,22 +60,45 @@ impl Interpreter {
...
 
-        for port_key in ctx.trace.port_map.keys() {
-            // We assume that there is only one `Instance` at the moment
-            let PortKey {
-                instance_id,
-                pin_id,
-            } = port_key;
+        // Map transaction parameters to their corresponding DUT pin values
+        // For INPUT parameters only - output parameters are populated later when observed
+        // For example: parameter 'a' -> value of 'DUT.a' from trace
```
ekiwi: @ngernest : why do you think that `a` maps to `DUT.a`?

ngernest: I think there are a couple issues with my implementation in this branch, I'll work on a rewrite over this weekend

ekiwi: OK. I guess there are still a couple of fundamental misunderstandings. Let's hope we can figure them out on Monday. In general, I would recommend trying to write as little code as possible. Then it will be easier for us to review it together.

ngernest:
Closing this branch and resuming work on #111 instead, which only contains the first few commits to this branch (improvements to pretty-printing)

PR #117, opened 2025-11-07
Statically check well-formedness of assignments, assertions & conditionals (`if` & `while`)

ngernest:
This PR implements the static checks for well-formedness of assignments / assertions / if-statements & while-loops that we discussed on Friday (11/7).

When reviewing this PR, I recommend first reading through `well-formedness.md`. 
This file documents the well-formedness constraints for these language features. 
These constraints are designed to make implementing the monitor tractable for programs with control flow (to be implemented in a future PR). 

Next, I recommend reading through `static_check.rs` -- this file contains most of the logic for implementing these checks. Specifically, this function contains functions that statically check well-formedness of assignments etc. These functions are invoked in `typecheck.rs`, since the type-checker has access to the individual exprs/statements whose well-formedness we are checking.

Afterwards, I recommend reading through the other Rust files that were changed & observing changes to the expected output of some Turnt / Snap tests. Note that as a result of these new well-formedness checks, a few Turnt tests now have well-formedness error messages as the protocols are now considered to be ill-formed (e.g. the condition for while-loops mention input parameters, which is no longer allowed). The expected output of these tests (+ the `.prot` files themselves) have been updated to reflect this. Additionally, the two `dual_identity_d1` Turnt tests have been merged into one, since `dual_identity_d1.prot` is now considered to be ill-formed under the new constraints (see comments in the file).

### DIFF
--- a/protocols/tests/snapshots/protocols__typecheck__tests__calyx_go_done_struct.snap
+++ b/protocols/tests/snapshots/protocols__typecheck__tests__calyx_go_done_struct.snap
@@ -8,10 +8,8 @@ warning: Inferred RHS type as u32 from LHS type u32.
 15 │   dut.ii := X;
    │   ^^^^^^^^^^^^ Inferred RHS type as u32 from LHS type u32.
 
-error: Expected oo to be a struct's input field,
-                but oo is not a field of a struct
+error: Cannot assign to function argument oo. Try using assert_eq if you want to check the value of a transaction output.
    ┌─ tests/calyx_go_done/calyx_go_done_struct.prot:16:3
    │
 16 │   oo := dut.oo;
-   │   ^^^^^^^^^^^^^ Expected oo to be a struct's input field,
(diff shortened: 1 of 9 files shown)


protocols/tests/snapshots/protocols__typecheck__tests__calyx_go_done_struct.snap
```diff
@@ -8,8 +8,10 @@ warning: Inferred RHS type as u32 from LHS type u32.
 15 │   dut.ii := X;
    │   ^^^^^^^^^^^^ Inferred RHS type as u32 from LHS type u32.
 
-error: Cannot assign to function argument. Try using assert_eq if you want to check the value of a transaction output.
+error: Expected oo to be a struct's input field,
+                but oo is not a field of a struct
```
ekiwi: I liked the old error message better? Can we bring it back?

Issue #124, opened 2025-11-11
[Monitor] Fix clock sampling + filter out `idle` transactions in WAL AXI Streaming example


ngernest:
The monitor currently produces the following output for the WAL AXI streaming example (#120):

```rust
idle()
idle()
...
idle()
send_data(0)
send_data(0)
idle()
...
send_data(1)
send_data(1)
idle()
...
```

This output exhibits two issues:

1. We need a way to filter out the `idle` transactions, perhaps by pursuing the `impl Idle` idea (#84). 

2. There are duplicate `send_data` calls in the monitor output. This is because the monitor currently assumes that each clock cycle in the waveform corresponds to a `step()` call, however this isn't the case with waveform files that aren't produced by our interpreter (e.g. the waveforms we get from WAL).

ngernest:
Marking this issue as closed as it has been resolved in #113, #127, #129 -- the remaining issues with the AXI example are discussed in issue #130

PR #127, opened 2025-11-14
[Monitor] Sample user-specified clock signal on rising edge + support AXI Streaming example

ngernest:
This PR adds an argument `--sample-posedge` to the monitor CLI, which allows the user to (optionally) specify a signal in the waveform to treat as a clock signal. If specified, this causes the monitor to sample the next rising edge of the clock signal when `WaveSignalTrace::step` is called. 

We demonstrate that this works by showing that the monitor supports the WAL Axis example waveform (which has been ported over to this PR from #120). 

To see this, run the following:
```bash
$ cd monitor/tests
$ cargo run --package protocols-monitor -- -p wal/advanced/axis.prot --wave wal/advanced/uart-axi.fst --instances uut_rx:AXIS --sample-posedge uut_rx.clk 
```

The following inferred trace is printed (see `monitor/tests/wal/advanced/axis.out` for the full trace):
```bash
idle()
...
send_data(1)
idle()
...
send_data(2)
idle()
...
send_data(3)
idle()
...
// Alternates between `send_data(n)` and lots of `idle()`'s for `n` from 3 up till 13
...
send_data(13)
...
send_data(16)
...
send_data(19)
```

Crucially, the duplicate occurrences of `send_data` mentioned in #124 have now been removed as a result of this PR. 

**Future Work**:
- Implement the `impl Idle` functionality, which will suppress `idle()` from the inferred trace above (making it easier for users to inspect the inferred transactions)
- Find more waveforms that aren't produced by Patronus, write example `.prot` programs for the corresponding protocol and see if the monitor works


monitor/src/signal_trace.rs
```diff
@@ -116,6 +135,9 @@ fn find_instances(
                     };
                     let waveform_bits = hierachy[var].length().expect("not a bit vector");
 
+                    // TODO: figure out how to set up `sample_posedge` to
+                    // refer to the clock signal (if one is specified)
+
```
ekiwi: You probably want to use: https://docs.rs/wellen/latest/wellen/struct.Hierarchy.html#method.lookup_var

You will need to split the name on `'.'`.

ngernest: thanks!

PR #140, opened 2025-12-09
[Monitor] AXI FIFO example from Brave New World artifact 

ngernest:
This PR examines whether the monitor works on two waveforms corresponding to buggy/fixed implementations of an AXI FIFO from the Brave New Artifact <https://github.com/ngernest/hardware-bugbase/tree/04250b9f01e5cf40e64a7d3bb271daf69f583107/d12-failure-to-update-frame-fifo>. (The corresponding Verilog and waveforms are included in this PR.)

The FIFO supports concurrent pop & push operations, which goes against the monitor's assumption that there is at most one transaction happening at each clock cycle. The bug manifests itself when there are concurrent transactions when the FIFO is full (i.e. after `push(6)` in the trace below, when the FIFO of capacity 4 has become full), but the monitor is unable to catch this. Specifically, on both waveforms, the monitor returns the same trace before it stops due to the same error:
```rust
push(1, ...)  // [time: 0ns -> 25ns] (thread 0)
push(1, ...)  // [time: 25ns -> 50ns] (thread 4)
push(1, ...)  // [time: 50ns -> 75ns] (thread 8)
push(2, ...)  // [time: 75ns -> 100ns] (thread 12)
push(3, ...)  // [time: 100ns -> 125ns] (thread 16)
push(4, ...)  // [time: 125ns -> 150ns] (thread 20)
push(5, ...)  // [time: 150ns -> 175ns] (thread 24)
push(6, ...)  // [time: 175ns -> 200ns] (thread 28)
pop(1, ...)  // [time: 200ns -> 225ns] (thread 33)
pop(2, ...)  // [time: 225ns -> 250ns] (thread 37)
push(6, ...)  // [time: 225ns -> 250ns] (thread 36)
Expected the no. of threads that started in cycle 9 & ended in cycle 11 to be at most 1, 
but instead there were 2 (["pop", "push"])
```

Here are the two waveforms (I have separated external vs internal signals in each of the Surfer screenshots). The bug occurs when `drop_frame_reg` is erroneously set to 0 in the bottom waveform, which causes `6` to be written to the FIFO even when the FIFO is full instead of being dropped. 

brave new world bugs

This PR also fixes a bug where incorrect time-stamps were displayed in the monitor output (previously, the [`Timescale.factor`](https://docs.rs/wellen/latest/wellen/struct.Timescale.html) field in Wellen was ignored when converting between time-units).

### DIFF
--- a/monitor/src/scheduler.rs
+++ b/monitor/src/scheduler.rs
@@ -2,7 +2,7 @@
 // released under MIT License
 // author: Ernest Ng <eyn5@cornell.edu>
 
-use anyhow::{anyhow, Context};
+use anyhow::{Context, anyhow};
 use baa::BitVecOps;
 use log::info;
 use protocols::{


ekiwi:
@ngernest : the next step here would be to define a `push_and_pop` protocol.

PR #142, opened 2025-12-10
[Monitor] Brave New World AXI FIFO Buffer Overflow Example 

ngernest:
This PR adapts the Protocols definitions in #140, applying the monitor to a [Buffer Overflow AXI FIFO bug in the Brave New World artifact](https://github.com/ngernest/hardware-bugbase/tree/macos-fixes/d4-buffer-overflow-frame-fifo). The buggy FIFO allows for buffer overflow, whereas the fixed FIFO prevents this. 

To distinguish the two traces, we have to add an extra `push_rejected` protocol definition that describes what happens when we attempt to push to the FIFO but the FIFO rejects it as it is full.

This is the diff between the two inferred traces produced by the monitor (obtained via [Patdiff](https://github.com/janestreet/patdiff)):
```bash
$ patdiff d4_buggy.out d4_fixed.out
------ d4_buggy.out
++++++ d4_fixed.out
@|-28,16 +28,17 ============================================================
 |push(6)  // [time: 92.5ns -> 95ns] (thread 224)
 |push(7)  // [time: 95ns -> 97.5ns] (thread 230)
 |push(8)  // [time: 97.5ns -> 100ns] (thread 236)
 |push(9)  // [time: 100ns -> 102.5ns] (thread 242)
 |push(233)  // [time: 102.5ns -> 105ns] (thread 248)
 |push(0)  // [time: 117.5ns -> 120ns] (thread 284)
 |push(1)  // [time: 120ns -> 122.5ns] (thread 290)
 |push(2)  // [time: 122.5ns -> 125ns] (thread 296)
 |push(3)  // [time: 125ns -> 127.5ns] (thread 302)
 |push(4)  // [time: 127.5ns -> 130ns] (thread 308)
 |push(5)  // [time: 130ns -> 132.5ns] (thread 314)
 |push(6)  // [time: 132.5ns -> 135ns] (thread 320)
 |push(7)  // [time: 135ns -> 137.5ns] (thread 326)
 |push(8)  // [time: 137.5ns -> 140ns] (thread 332)
 |push(9)  // [time: 140ns -> 142.5ns] (thread 338)
 |push(233)  // [time: 142.5ns -> 145ns] (thread 344)
+|push_rejected(0)  // [time: 157.5ns -> 160ns] (thread 379)
```

The monitor infers an extra `push_rejected(0)` transaction for the *fixed* protocol, which occurs as we try to `push` an extra value `0` to the FIFO, but the FIFO (correctly) rejects this push as it is already full (and the fix in the Verilog prevents buffer overflow).

Note: on the fixed waveform, the monitor exits with the following message:
```bash
Thread 379 (`push_rejected`) finished but there are other threads with the 
same start time (157.5ns) in the `next` queue, namely ["push"]
Error: Monitor failed
```

This message is expected, since when `push_rejected` finishes successfully and we reach the end of the waveform, the `push` transaction is stuck in its while-loop, waiting for `ready` to become 1 and calling `step` on each loop iteration, so the `push` thread remains in the `next` queue when the monitor finishes:
```rust
fn push<DUT: AxisFifo>(in data: u8) {
     ...
    // Wait for FIFO to be ready to accept data
    while (!(DUT.s_axis_tready == 1'b1)) {
        step();
    }
   ...
}
```

I'm not sure if we should accept this limitation in the monitor, or if we should modify the `push` protocol definition to impose some sort of "timeout".

***

Other (housekeeping) changes:
- Renamed folders containing Brave New World artifacts to follow Kevin's naming conventions in the RTL-Repair artifact
- All Brave New World artifact examples are in the `monitor/tests/fpga-debugging` subfolder.
- The key files in this PR are in the `fpga-debugging/axis-fifo-d4`
- I also moved Francis's Brave New World (synthetic) examples to a separate folder called `brave_new_world_francis`

### DIFF
--- a/monitor/tests/fpga-debugging/axis-fifo-d12/README.md
+++ b/monitor/tests/fpga-debugging/axis-fifo-d12/README.md
@@ -1,0 +1,65 @@
+# D12 - Failure-to-Update - Frame FIFO
+
+This README was taken from: https://github.com/efeslab/hardware-bugbase/tree/bugs/d12-failure-to-update-frame-fifo
+
+**Source:** Verilog-axis(Verilog AXI Stream Components): https://github.com/alexforencich/verilog-axis/commit/a9c04a465150ae5fc9cd8f32906213d9ba9afb06
+
+This bug is caused by failing to keep value in the register `drop_frame_reg`. Because the packets must be dropped at frame granularity, `drop_frame_next` should have the default value of `drop_frame_reg` at the beginning of the following always block.
+
+
+When an axis-FIFO buffer is full, it will stop writing new packets from the input.
+
+`drop_frame_reg` will be turned on and the FIFO should stop storing anything from the input until the `tlast` is set on (the current frame is finishing).
+
+So `drop_frame_reg` should be continuously be 1 until `tlast` rises.
+The value should be “kept” in reg.
+
+### Synthetic Code
+```verilog
+// Write logic
+always @* begin
+    write = 1'b0;
+
+    drop_frame_next = 1'b0;
+    overflow_next = 1'b0;
+    bad_frame_next = 1'b0;
+    good_frame_next = 1'b0;
+
+    wr_ptr_next = wr_ptr_reg;
+    wr_ptr_cur_next = wr_ptr_cur_reg;
+
+    if (s_axis_tready && s_axis_tvalid) begin
+        // transfer in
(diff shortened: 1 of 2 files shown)


ekiwi:
Thanks! I would expect the error message to be a bit different though. Let's talk about that tomorrow

PR #146, opened 2025-12-11
[Monitor] Python scripts for benchmarking via Hyperfine (included in CI)

ngernest:
This PR creates three Python scripts for benchmarking the monitor (all in the `scripts` sub-directory):
- `benchmark_monitor.py` measures the execution time (CPU wall-clock time) of the monitor on all existing test files, averaged over 10 runs for each file and outputs results in a CSV (within the `benchmark_results` sub-directory)
- `plot_benchmark_results.py` takes the CSV and creates a scatter plot
- `check_performance_regression.py` examines the CSV results and compares them to a baseline CSV to see if there are any performance regressions (increase in mean execution time of over 10%) in the monitor 
  + if there is no baseline result CSV, the script uses the CSV from the most recent run as the baseline
  + Note: this means that the CSVs will need to be checked into Git 

Here's the plot that is produced:
benchmark_plot

This PR also sets up a `uv` project to manage Python dependencies (this will become useful as we possibly explore implementing Python bindings in the future). 

Note: the scripts should be run from the top-level directory as follows:
```bash
$ uv run scripts/benchmark_monitor.py
$ uv run scripts/plot_benchmark_results.py
$ uv run scripts/check_performance_regression.py
```

The first time you do `uv run`, `uv` will create a new Python virtual environment with the necessary dependencies based on the `uv.lock` and `pyproject.toml` files in the repo (part of this PR).

### DIFF
--- a/scripts/benchmark_monitor.py
+++ b/scripts/benchmark_monitor.py
@@ -43,8 +43,8 @@ def main():
             "--export-json", "tmp_hyperfine.json",
             # 3 warmup runs to run benchmarks on a warm cache
             "--warmup", "3",
-            # Minimal output for CI
-            "--style", "basic",
+            # Suppress hyperfine output (results are checked separately)
+            "--style", "none",
             f"turnt --env monitor {pf}",
         ]
 


ekiwi:
@ngernest before we merge this, it would be great if you could setup a new CI task that runs some of these Python scripts.

PR #147, opened 2025-12-12
[Monitor] Normalize time by no. of `step`s in benchmarks + Run Turnt tests in parallel in CI

ngernest:
1. When benchmarking the monitor, we divide the wall-clock time by the no. of logical `step`s taken. An extra CLI flag `--print-num-steps` was added to the monitor to print the total no. of steps taken in a separate `.prof` (short for "profiling") file, which is then parsed by the Python benchmark scripts. 

The benchmarking script was also updated to avoid measurement overhead from Turnt / avoid statistical outliers, to run the monitor executable using `cargo run --release`, and also include a progress bar (via the `tqdm` library).

Here are the benchmarks:
benchmark_plot

- To support this, a separate Turnt environment for benchmarking the monitor was created.
- Most of the files changed are just creating extra `.prof` files for the monitor benchmarks + creating empty `.err` files for test cases that aren't expect to fail (both of these required by Turnt).

3. Updated the Turnt tests in the Justfile + CI to run in parallel (by using Turnt's `--parallel` flag) for efficiency


ngernest:
@ekiwi Updated the scripts to plot the inverse (no. of steps per second):

benchmark_plot

Issue #143, opened 2025-12-11
[Monitor] Support multiple structs in the same Protocols file


ngernest:
For the Brave New World AXI FIFO examples, it's helpful to have two structs in the `*.prot` file called `Sender` and `Receiver`, that support different "methods":

```rust
// Only `Sender` supports `push`
struct Sender {...}
fn push<DUT: Sender> {...}
fn idle<DUT: Sender> {...}
fn reset<DUT: Sender> {...}

// Only `Receiver` supports `pop`
struct Receiver{...}
fn pop<DUT: Receiver> {...}
fn idle<DUT: Receiver> {...}
```

Intuitively, `Sender` and `Receiver` distinguishes transactions that provide an *input* to the FIFO, vs transactions that produce an *output*.

Remarks:
- The implementations of `idle` for `Sender` and `Receiver` are different since they set different signals (`valid` vs `ready` respectively) to 0.
- `reset` is the same for both `Sender` and `Receiver` -- without loss of generality, we only implement it for `Sender`.

To support this in the monitor implementation, we need to do the following:
- Have `n` schedulers, where each scheduler contains the 4 queues (`current, next, finished, failed`), where `n` = no. of structs in the Protocols file (in the example above, we'd have `n = 2` schedulers for `Sender` & `Receiver`)
- Run each scheduler until the next `step`
- Synchronize the schedulers for each struct, and only call `trace.step()` when all the schedulers are the same 
- Run each of the schedulers in a round-robin fashion 
- Challenge: how do you figure out when a scheduler is done
- Own the trace in a function that runs the scheduler, then have a reference to the Scheduler struct 
- When printing out inferred transactions, you also need to print out the name of the struct for disambiguation purposes, e.g. 

```rust
Sender.push(3)  // 100ns
Receiver.pop(2)  // 105ns
```

PR #148, opened 2025-12-18
[Monitor] Support multiple structs (each w/ their own scheduler) + handle BNW `axis-async-fifo-c4` bug

ngernest:
This PR addresses #143 & supports the Brave New World asynchronous FIFO bug `axis-async-fifo-c4` (#145).

In the Protocols file, there are now two structs `Sender` and `Receiver`, which distinguish transactions that provide an input to the FIFO, vs transactions that produce an output.

```rust
// Only `Sender` supports `push`
struct Sender {...}
fn push<DUT: Sender> {...}
fn idle<DUT: Sender> {...}
fn reset<DUT: Sender> {...}

// Only `Receiver` supports `pop`
struct Receiver{...}
fn pop<DUT: Receiver> {...}
fn idle<DUT: Receiver> {...}
```

Intuitively, Sender and Receiver distinguishes transactions that provide an input to the FIFO, vs transactions that produce an output. 

To support multiple structs in the monitor, we create *multiple* schedulers (one scheduler per struct), and we create a `GlobalScheduler` that coordinates between each individual scheduler.

Key files changed:
- `scheduler.rs`
- `global_scheduler.rs`
- `global_context.rs`
   - Some of the changes involve moving where the `WaveSignalTrace` is stored (since it doesn't implement `Clone`), and making functions explicitly take a reference to the trace as an additional argument so that we avoid issues with the borrow checker

From the [description of the bug in the BNW artifact](https://github.com/ngernest/hardware-bugbase/tree/04250b9f01e5cf40e64a7d3bb271daf69f583107/c4-signal-asynchrony-axi-stream-fifo), this is a signal asynchrony bug. Specifically, in the buggy implementation, data continues to be accepted by the FIFO during a multi-cycle reset transaction (`ready` is erroneously set to 1), whereas this does not happen in the fixed version.

Here is the diff in the transaction-level trace inferred by the monitor for the two waveforms
(left: buggy, right: fixed):

Screenshot 2025-12-19 at 1 38 59 PM

This is how the diff illustrates the bug:
- In the buggy waveform, the `Receiver::pop(2)` starts at 25ns, immediately after the reset. 
- In the fixed waveform, we see multiple `Receiver::idle()` transactions after the reset (25ns-75ns), and the first `Receiver::pop()` transaction doesn't start until 100ns, indicating the FIFO correctly waits for the reset to fully complete before accepting transactions. 

The benchmark results CSV has also been updated in this PR due to the changes in the monitor code.

**Future work**
- Thread IDs are currently only unique within one single scheduler, but they need to be made unique across *all* schedulers. (See #149) 
    + Right now, this doesn't cause any functional problems, but making all thread IDs unique will make the logs more consistent

### DIFF
--- a/monitor/src/main.rs
+++ b/monitor/src/main.rs
@@ -85,7 +85,6 @@ struct Cli {
     print_num_steps: bool,
 }
 
-#[allow(unused_variables)]
 fn main() -> anyhow::Result<()> {
     // Parse CLI args
     let cli = Cli::parse();
@@ -179,7 +178,7 @@ fn main() -> anyhow::Result<()> {
     // Multi-struct mode: create a GlobalScheduler with one scheduler per design
     let mut schedulers = vec![];
 
-    for (inst_id, design) in dut_designs.into_iter().enumerate() {
+    for design in dut_designs.into_iter() {
         // Filter transactions that belong to this design
         let design_transactions: Vec<(Transaction, SymbolTable)> = design
             .transaction_ids
--- a/monitor/src/scheduler.rs
+++ b/monitor/src/scheduler.rs
@@ -15,13 +15,12 @@ use rustc_hash::FxHashSet;
 use crate::{
     global_context::GlobalContext,
     interpreter::Interpreter,
-    signal_trace::{SignalTrace, StepResult, WaveSignalTrace},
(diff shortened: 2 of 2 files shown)


ekiwi (approved):
Looks good. Feel free to merge.

Issue #149, opened 2025-12-19
[Monitor] Make thread IDs unique across all structs


ngernest:
In #148 (draft PR), thread IDs are unique for all threads corresponding to the same scheduler (one scheduler per struct). However, in a multi-struct setting, we can have the same thread ID corresponding to threads for different schedulers (i.e. different structs). Right now, this doesn't cause any issues, since threads from different schedulers never interact with each other, but for the sake of consistency in the logs + error messages, we should make thread IDs unique across all structs. 

In terms of the implementation, at scheduler-creation time, we specify that all new threads created must have their IDs start from some base value, and we update this base value accordingly for different schedulers, ensuring that thread IDs for different schedulers reside in different (disjoint) integer intervals.

PR #150, opened 2025-12-23
[Monitor] Make thread IDs unique across all schedulers in multi-struct mode

ngernest:
Addresses issue #149. When there are multiple structs (i.e. multiple schedulers), we print the name of the struct (scheduler) as a prefix before printing the thread ID in the logs. 

In the diff below, the logs in the right (the updated logs) contain the struct name as a prefix:
Screenshot 2025-12-22 at 8 42 27 PM


Nikil-Shyamsunder:
this lgtm

PR #154, opened 2026-01-10
[interp] Relax requirement that protocols have to end with fork(); step() 

Nikil-Shyamsunder:
This PR makes fork() implicit at the end of protocol execution. Previously, protocols were required to explicitly call fork() exactly once, and many protocols ended with a boilerplate pattern of fork(); step(); where the final cycle contained only don't-care assignments and the fork call. Now, protocols simply end with step(), and if they haven't forked yet, the scheduler automatically forks and sets all input pins to DontCare.

Interpreter/Scheduler Changes:
We already had all the plumbing for this from a fixed point implementation with implicit forking I had before we wrote the well-formedness checks. I didnt even remove this logic, so very little needs to change here. I just removed the finished_without_fork error from the scheduler. When a thread completes execution with step() as its final statement, it no longer errors if fork() wasn't called. Instead, the existing implicit fork logic at the end of run_thread_until_next_step handles spawning the next transaction, and reset_all_input_pins() already sets all inputs to DontCare when threads finish.

Protocol Changes:
Updated 8 protocol functions across 5 files where the final cycle contained only fork() and/or X assignments: push, idle, and reset in fifo.prot; identity in dual_identity_d0.prot; pcpi_mul and pcpi_mul_reset in pcpi_mul.prot; pcpi_mul in pcpi_mul_no_reset.prot; and the main function in serv_regfile.prot. Protocols that had meaningful work in the final cycle (assertions, concrete assignments) were left unchanged.

Snapshot Changes:
In theory, the outputs should be identical since this is just a QoL syntax change. But i had to update three snapshot files in protocols/tests/fifo/ (pop_before_push_fail.out, pop_empty_fail.out, push_pop_conflict.out). These changes are purely cosmetic: the line numbers in error messages shifted because we removed lines from fifo.prot. The actual error messages and semantics remain identical.

### DIFF
--- a/protocols/src/scheduler.rs
+++ b/protocols/src/scheduler.rs
@@ -473,19 +473,9 @@ impl<'a> Scheduler<'a> {
 
                     // Check if the last executed statement was `step()`
                     if let Stmt::Step = thread.todo.tr[current_stmt_id] {
-                        if forks_enabled && !thread.has_forked {
-                            // Throw an error if forks are enabled but the
-                            // thread finished without making any calls to `fork()`
-                            info!(
-                                "  ERROR: thread did not make any calls to `fork()`, terminating thread"
-                            );
-                            let error =
-                                ExecutionError::finished_without_fork(thread_id, transaction_name);
-                            self.results[thread_id] = Err(error);
-                        } else {
-                            // Thread completed execution successfully
-                            info!("  Execution complete, no more statements.");
-                        }
+                        // Thread completed execution successfully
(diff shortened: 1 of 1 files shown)


ngernest (approved):
Thanks! Can you point out where in the current code new threads are initialized with all input pins set to `DontCare`? I tried looking through this branch and wasn't able to find it. 

Otherwise, this looks good to me! I am impressed at how few code changes were required while still making everything work.

PR #155, opened 2026-01-11
[Interp] Implicit input reapplication

Nikil-Shyamsunder:
Now that we've removed fixed-point iteration, this PR implements implicit re-application of inputs between
cycles. The core idea is that inputs should persist unless explicitly changed; we track the most recent
assignment to each input per transaction, then re-apply them as NewValues at cycle boundaries.

Consider this example:

```
fn add<DUT: Adder>(in a: u32, in b: u32, out s: u32) {
DUT.a := a;
DUT.b := b;
step();
fork();
// not setting input pins to DontCare!
step();
assert_eq(s, DUT.s);
step();
}
```

Without implicit re-application, omitting `DUT.a := X; DUT.b := X;` in cycle 2 would cause values `a` and `b`
to persist as OldValues, silently allowing the next pipelined transaction to overwrite them without error.
With implicit re-application, we correctly get a ConflictingAssignment error unless we explicitly release
with `X`. This ensures protocol writers must be explicit about when inputs should be released for reuse by
concurrent transactions.

Cases involving combinational dependencies are treated exactly the same as if the assignments were explicit. Consider:

```
struct Passthrough {
in a: u1,
out b: u1,  // b = a combinationally
}

fn foo<DUT: Passthrough>() {
DUT.a := 0;
step();

if (DUT.b == 1) {
DUT.a := 1;
} // implicit: else { DUT.a := 0; }
step();
}
```

  In cycle 2, the transaction observes DUT.b (which forbids future assignments to a due to combinational
  dependency tracking). Since b = 0, we don't enter the if-branch. Now implicit re-application wants to
  re-assert DUT.a := 0, but a is forbidden. This is an error: if a transaction observes an output that is
  combinationally dependent on input i, and i has a held value that would be re-applied, we report the
  same "ASSIGNED FORBIDDEN PORT" error as an explicit assignment would produce. This forces protocol writers
  to explicitly set inputs to DontCare before observing combinationally dependent outputs.

## Implementation

The interpreter now maintains a `thread_sticky_inputs` map for each thread, tracking the most recent assignment to each input pin. At the end of each thread's execution within a cycle, `reapply_held_inputs()` iterates through
held values: if the input is in `forbidden_inputs` (due to observing a combinationally dependent output),
it's forced to DontCare; otherwise the held value is re-applied as a NewValue. When a thread completes and
implicitly forks a new transaction, we reset all input pins before the fork so the new thread starts with
clean state.

When I was implementing the passthrough example from above, I also found a bug with how empty if/else blocks are handled. The interpreter was returned `None` as the next statement on seeing an empty if block, when really it should be returning the next statement after the if/else chain from the `next_stmt_mapping`, which is what we did. 

## Tests

Added `no_dontcare_conflict.prot` and `.tx` for adder_d2, which is described above. Added the passthrough example depicted above.

Modified `alu_d1.prot` and `alu_d2.prot` to add explicit `DUT.op := X` assignments in cycles where the opcode
should be released, as required by the new semantics.

### DIFF
--- a/protocols/src/interpreter.rs
+++ b/protocols/src/interpreter.rs
@@ -740,26 +740,25 @@ impl<'a> Evaluator<'a> {
     ///
     /// For each input with a held concrete value:
     /// - If the input is in `forbidden_inputs` (we observed a comb-dependent output),
-    ///   force it to DontCare instead of re-applying the held value.
+    ///   return an error (same as explicit assignment to a forbidden input).
     /// - Otherwise, re-apply the held value as a NewValue.
     ///
-    /// Returns an error if re-application causes a conflicting assignment.
+    /// Returns an error if re-application would assign to a forbidden input or cause a
+    /// conflicting assignment.
     pub fn reapply_held_inputs(&mut self) -> ExecutionResult<()> {
         for (symbol_id, held_value) in self.thread_sticky_inputs.clone().iter() {
             if let Some(value) = held_value {
                 // Check if this input is forbidden due to observing a comb-dependent output
                 if self.forbidden_inputs.contains(symbol_id) {
-                    // TODO: Could error here instead of forcing DontCare?
-                    info!(
-                        "Forcing input '{}' to DontCare due to combinational dependency",
-                        self.st[symbol_id].name()
-                    );
-                    if let Some(input_val) = self.sim_input_vals.get_mut(symbol_id) {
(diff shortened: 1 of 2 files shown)


ekiwi:
@Nikil-Shyamsunder: Do you plan to update this PR after our conversation yesterday, or is it still good to go in its current state?

Issue #167, opened 2026-01-30
Add non-deterministic loop or for-loop with fixed no. of iterations


ngernest:
Consider the Brave New World bug in #158. There should be a way for us to vary the no. of cycles in which we apply back pressure. 

There are two options:
1. **Non-deterministic loop operator** (akin to Kleene star)

Here, we non-deterministically choose the no. of loop iterations. Example:

```rust
fn recv<DUT: AXISManager>(
    out data: u32,
    out is_last: u1,
) {
    DUT.i_aresetn := 1'b1;         // Keep out of reset

    // we are free to signal ready or not, it does not matter until valid is asserted
    DUT.i_tready := X;


    // wait for valid data
    while(DUT.i_tvalid == 1'b0) {
        step();
    }

    // now we are allowed to stall for an arbitrary number of cycles as long
    // as we keep ready low
    DUT.i_tready := 1'b0;
    while(X) {
        // valid has to remain true even if we are stalling the transfer!
        assert_eq(DUT.i_tvalid, 1'b1);
        assert_eq(DUT.i_tdata, data);
        step();
    }
    // set ready to true for the final cycle
    DUT.i_tready := 1'b1;
    assert_eq(DUT.i_tvalid, 1'b1);
    assert_eq(DUT.i_tdata, data);

    // One cycle for the transfer to complete
    step();
}
```

This is challenging for users of the interpreter and the monitor since they have no way to control the no. of iterations in which back pressure is applied. This motivates the second approach:

2. **For-loop with a fixed no. of iterations**

Here, we have an extra input parameter which fixes the no. of loop-iterating. The user of the interpreter needs to decide ahead of time the no. of loop iterations. On the other hand, the monitor needs to *infer* the value for this input parameter by forking two threads at each loop iteration (one thread executes the loop body while the other thread exits the loop) and seeing which thread succeeds. This would make the monitor implementation more complex (but would be simple to implement in the interpreter). 

Example:
```rust
fn foo(..., in back_pressure_cycles: u8) { 
   ...
   for back_pressure_cycles iterations {
     ...
   }
}
```

Another potential issue in the monitor is if two threads succeeds, one which continues the loop body and one which exits the loop. In this case, we need a way to break ties. Kevin mentioned letting the thread which continues the loop body succeed since it allows for a more compact transaction trace (e.g. a fully combinational adder which maintains its inputs/outputs for 3 cycles, as opposed to 3 separate add transactions).

PR #171, opened 2026-02-06
[Interp] Extend language with `repeat ... iterations` bounded loop

ngernest:
This PR addresses #167 (see Slack) for the interpreter by extending the language with a `repeat num_iters iterations { ... }` bounded loop construct, which executes its loop body for _exactly_ `num_iters` iterations (`num_iters` must be an input parameter for the enclosing protocol). Example:

```rust
fn foo<...>(..., in num_cycles: u8) { 
   // Bounded loop that runs exactly for `num_cycles` iterations
   repeat num_cycles iterations {
     ...
   }
}
```

A future PR will extend the monitor to support `repeat` loops (since more changes are required in the monitor codebase).

**Tests**:
For these tests, I checked the waveform files produced by the interpreter to make sure the right no. of iterations are being executed.

- `busy_wait.prot` : Tests that busy-waiting (i.e. `repeat num_iters iterations { step() }` doesn't change the result of an adder (since concrete assignments persist)
- `nested_busy_wait.prot`: same as above, but with nested `repeat` loops
- `loop_with_assigns.prot`: Tests that driving values onto input ports in the body of a `repeat` loop works
- `fifo_bounded_loop`: Tests that `push`-ing the same value for `n` times and subsequently `pop`-ping the same value for `n` times returns an empty FIFO (`push_pop_loop_empty.tx`), and that if the no. of times we `push` / `pop` is different, the resultant FIFO is non-empty (`push_pop_loop_non_empty.tx`)
- Some extra parser / lexer / serializer tests to make sure `repeat` loops are parsed/typechecked properly

### DIFF
--- a/protocols/src/interpreter.rs
+++ b/protocols/src/interpreter.rs
@@ -117,6 +117,9 @@ pub struct Evaluator<'a> {
 
     /// Random number generator used for generating random values for `DontCare`
     rng: StdRng,
+
+    // Maps a `StmtId` pair to the no. of iterations remaining for that particular loop.
+    bounded_loop_remaining_iters: FxHashMap<StmtId, u128>,
 }
 
 impl<'a> Evaluator<'a> {
@@ -264,6 +267,7 @@ impl<'a> Evaluator<'a> {
             per_thread_input_vals,
             current_todo_idx: 0,
             assertions_enabled: false,
+            bounded_loop_remaining_iters: FxHashMap::default(),
             rng,
         }
     }
@@ -292,6 +296,7 @@ impl<'a> Evaluator<'a> {
         self.args_mapping = Evaluator::generate_args_mapping(self.st, todo.args);
         self.next_stmt_map = todo.next_stmt_map;
         self.current_todo_idx = todo_idx;
+        self.bounded_loop_remaining_iters = todo.bounded_loop_remaining_iters;
 
         // Clear forbidden ports when switching to a new thread context
         self.forbidden_inputs.clear();
@@ -484,6 +489,11 @@ impl<'a> Evaluator<'a> {
     /// Used by the scheduler to check for conflicting assignments across threads.
     pub fn per_thread_input_vals(&self) -> &FxHashMap<SymbolId, PerThreadValues> {
         &self.per_thread_input_vals
+    }
+
+    /// Returns a clone of the bounded loop remaining iterations map.
+    pub fn bounded_loop_remaining_iters(&self) -> FxHashMap<StmtId, u128> {
+        self.bounded_loop_remaining_iters.clone()
     }
 
     /// Get the next statement after the given statement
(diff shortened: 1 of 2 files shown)


protocols/tests/adders/adder_d1/busy_wait.prot
```diff
@@ -0,0 +1,24 @@
...
+// Uses a bounded loop that busy-waits for `num_iters` cycles.
+// Since concrete assignments to DUT input ports persist, 
+// calling `step()` multiple times in the loop with the same inputs 
+// should still produce the correct result on the adder. (This allows us to 
+// reuse the same Verilog file as the other `adder_d1` tests.)
+fn add_busy_wait<DUT: Adder>(in a: u32, in b: u32, in num_iters: u8, out s: u32) {
+  DUT.a := a;
+  DUT.b := b;
+  repeat num_iters iterations {
+    step();
```
ekiwi: Would it make sense to add an `assert_eq(s, DUT.s)` here?

ngernest: Thanks, good point! Changed it to:

```rust
...
repeat num_iters iterations {
    step();
    assert_eq(s, DUT.s);
}
...
```

Nikil-Shyamsunder:
I don't see any changes to the well-formedness checks for this... is this a special case where we allow control flow to be determined by the value of an input port directly?

ngernest:
@Nikil-Shyamsunder Yep! I think this `repeat ...` loop construct is a special case that's separate from `while` loops (so the well-formedness checks for `while` loops don't apply to `repeat` loops).

protocols/src/interpreter.rs
```diff
@@ -103,8 +107,19 @@ pub struct Evaluator<'a> {
...
+
+    /// Maps a `(TodoIdx, StmtId)` pair (representing a protocol that is
+    /// executing a particular bounded loop) to the no. of iterations remaining
+    /// for that particular loop.              
+    /// - **Invariant**: the no. of remaining iterations is always non-zero
+    ///   (if it reaches zero, we remove the entry from this map).       
+    /// - In the key, we need the `StmtId` to allow for nested loops,
+    ///   and we also need the `TodoIdx`, since the same `StmtId` could be active
+    ///   in differnt threads.
+    bounded_loop_remaining_iters: FxHashMap<(TodoIdx, StmtId), u64>,
```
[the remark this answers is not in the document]
Nikil-Shyamsunder: I liked the change, but I made a minor switch to it. I moved the data into the `Todo` struct instead. That way we guarantee every todo is completely isolated (we never have the evaluator have access to multiple Todo's data at once). Neither solution is too elegant because of some structural issues in the interpreter that might deserve a rewrite one day, but I think this minor edit is slightly cleaner.

Issue #172, opened 2026-02-09
[Monitor] Perform BFS over possible protocol traces and show all possible ambiguous traces


ngernest:
Right now, when the monitor encounters `fork()`, it appends one new thread for each possible protocol to the scheduler's ready queue. That is, if there are `k` possible protocols, then `k` new threads are appended to the queue every time we encounter `fork()`. 

The issue with this design is that it conflates these two things:
1. The "protocol trace state space" (i.e. the possibilities of different protocols ending up in the final trace)
2. The scheduling policy (i.e. the concurrent behavior of multiple threads)

Instead, we want to refactor the monitor so that when it encounters `fork()`, we "clone" the entire `Scheduler` state, except with exactly *one* possible protocol candidate for the forked-off thread appended to the queue. We do this for every possible protocol. That is, if there are `k` possible protocols, then we branch off into `k` different `Scheduler` clones, whose queues are all the same except for the final element (the thread for the protocol candidate). 

Whiteboard sketch from 2/9 meeting:
![Image](https://github.com/user-attachments/assets/b4514726-c430-452c-a27d-e8ea7cfb25b0)

This feature is implemented in anticipation of supporting `repeat` loops (#171) in the monitor, 
but as a first pass, we should aim to preserve all existing monitor functionality + tests after this refactoring (i.e. don't support `repeat` loops in the monitor just yet). Notably, the monitor spawns a new instance every time it proceeds to a new loop iteration and decides whether to run the loop body / exit the loop. 
(Another approach would be to spawn `N` new instances at when the monitor first encounters a `repeat N iterations { ... }` statement, with each iteration running a different no. of iterations. However, this approach is infeasible since `N` could be arbitrarily large.)

Also, the current monitor implementation enforces the current constraint: "out of all the threads that were launched in the same cycle, at most one of them can succeed" (see `validate_finished_and_failed_threads` in `scheduler.rs`). This ensures that the monitor can always report exactly one protocol trace at the end. 

However, when the monitor supports `repeat` loops in the future, we will need to relax this restriction, 
since the same protocol could match with different values of `N` (where `N` is the no. of iterations for a `repeat` loop). Tie-breaking heuristics (e.g. always preferring the protocol that matches with the highest value of `N`) are finicky in practice in the presence of nested loops (maximizing one value could in theory cause another value to be minimized) and the same `N` being used for multiple loops (e.g., `repeat N iterations { ... }; ...; repeat N iterations { ... }`). 

[...]

PR #174, opened 2026-02-10
[Monitor] Refactor: clone scheduler when forking new protocols

ngernest:
This PR addresses #172 for the existing monitor. That is, instead of appending a new `Thread` for each candidate protocol every time we encounter `fork`, we instead clone the entire `Scheduler` and perform BFS over the possible transaction-level traces (see #172 for more details). This refactor is done in anticipation of supporting `repeat ...` loops in the monitor (which will be handled in a future PR). 

As mentioned in #172, as part of this PR, the monitor no longer enforces the constraint that "out of all the threads that were launched in the same cycle, at most one of them can succeed", i.e. the monitor can now report *several* candidate traces at the end.

To avoid `scheduler.rs` being too large, I've created two new files `queue.rs` and `types.rs`, which contain all queue-related operations / type definitions relevant to the monitor. 

All tests pass, with the exception that the monitor now reports several traces for the Brave New World `axi-stream-s2` bug. For example, `s2_fixed` now has 3 traces which are all identical except for the no. of consecutive `stall` transactions:
  - Trace 0: `...recv(6), stall(7,0), stall(7,0), reset(), reset()` (2 stalls)
  - Trace 1: .`..recv(6), stall(7,0), reset(), reset()` (1 stall)
  - Trace 2: `...recv(6), reset(), reset()` (0 stalls)

Other remarks:
- The difference for the `axi-adapter-s3` output (another BNW AXI-Stream bug) is that the final `recv` transaction has now been replaced with `recv_last` (I added an extra `assert_eq` in the protocol to better distinguish `recv` from `recv_last`, this was previously missing) 
- I believe all other differences in Turnt test outputs are just individual thread IDs (as opposed to different traces being reported)

Note: There is some overhead associated with cloning the `Scheduler` every time a `fork` is encountered, which is why I updated the CSVs containing performance benchmarks in this PR in order to get the `Check Performance Regressions` CI check to pass. (The CI check was previously comparing the CSVs with the ones on the `main` branch, which are outdated  -- I updated it in this PR to only compare the CSVs on the local branch.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,9 +89,6 @@ jobs:
           sudo dpkg -i hyperfine_1.19.0_amd64.deb
       - name: Build monitor (release)
         run: cargo build --release --package protocols-monitor
-      - name: Restore baseline benchmark results
-        run: |
-          git show origin/main:benchmark_results/benchmark_results.csv > benchmark_results/baseline_results.csv 2>/dev/null || echo "No baseline found"
       - name: Run benchmarks
         run: uv run scripts/benchmark_monitor.py
       - name: Check for performance regressions


ngernest:
Yes! 

> What about the protocols allows for this ambiguity?

Here is the relevant section of the waveform (`s2_fixed.fst`), the key section are the two cycles from 1012.5 ns - 1062.5 ns:

Screenshot 2026-02-12 at 1 41 35 PM

During these two cycles, we have both `ready = 0` and `valid = 1`. This causes both the `stall` and `idle` transactions to match, since `stall` requires both `ready = 0` and `valid = 1`, whereas `idle` only requires `ready = 0` and doesn't say anything about `valid` (i.e. `valid` is `DontCare` since no concrete assignment has been performed to `valid`). (The definitions of these protocols are copied below)

Since `stall` and `idle` both only last for one cycle but the aforementioned behavior lasts for two cycles, there are 3 possibilities:
1. `stall` both cycles
2. `stall` the 1st cycle, then `idle` the 2nd cycle
3. `idle` both cycles
(There is actually a 4th possibility, which is `idle` the 1st cycle and `stall` the 2nd cycle. The check for deduplicating traces in `global_scheduler.rs` strips all protocols annotated with `#[idle]` from the trace and then compares them, so after stripping `idle`s, the 4th possibility and the 2nd possibility are considered the "same", so only 3 possible traces are displayed to the user.)

To fix this, we can add an extra `assert_eq(DUT.valid, 0)` to `idle` to distinguish it from `stall`.  

```rust
// From s2_fixed.prot

// STALL: Assert backpressure (i_tready=0) while manager has valid data
// AXI-Stream requires output args (tdata, tlast) to remain stable during stall
// Only matches when the waveform contains i_tvalid=1 (i.e. there is some valid data and it's available to stall)
prot stall<DUT: AXISManager>(out data: u32, out last: u1) {
    DUT.i_aresetn := 1'b1;         // Keep out of reset
    DUT.i_tready := 1'b0;          // Apply backpressure

    // Stall only applies when manager has valid data
    // If i_tvalid=0, this fails and we instead have an `idle` transaction
    assert_eq(DUT.i_tvalid, 1'b1);

    // Capture output values before the stall cycle
    assert_eq(DUT.i_tdata, data);
    assert_eq(DUT.i_tlast, last);

    step();

    // Verify outputs remained stable during the stall
    assert_eq(DUT.i_tdata, data);
    assert_eq(DUT.i_tlast, last);
}

// IDLE: No transaction - i_tready is deasserted
#[idle]
prot idle<DUT: AXISManager>() {
    DUT.i_aresetn := 1'b1;         // Keep out of reset
    DUT.i_tready := 1'b0;
    // <--- Missing `assert_eq(DUT.valid, 1'b0)` here, if we added it we could distinguish it from `stall`
    step();
}
```

Issue #128, opened 2025-11-19
Extend language with `prot fn` keyword for marking top-level entry points (protocols)


ngernest:
(Creating an issue for this discussion just so we don't forget)

Paraphrasing our [Slack discussion on 11/3](https://cucapra.slack.com/archives/C0814JN4VMH/p1762186575860769): 

Adrian suggested adding an additional keyword (`prot`) to the DSL for marking the `fn` definitions that are top-level entry points, which make them protocols? That is, the user would write `prot fn foo` to declare a top-level protocol `foo`. 

Nikil raised the point that in the future, if we want to allow protocols to call other protocols as a method of composition, we may want to define some "sub-protocols" that are correct when used within the context of the larger parent protocol but incorrect to run as an independent transaction on their own. To support this kind of composition, adding the `prot` keyword to the language allows us to (syntactically) distinguish sub-protocols that can only be used in composition with others, versus "true" protocols (those the user can call in a `.tx` file).

PR #176, opened 2026-02-11
Chore: Update parser to require `prot` instead of `fn` for protocol definitions

ngernest:
We've manually replaced `fn` with `prot` in code-snippets in various research proposals we've written for this project (e.g. Qualcomm / NSF), so it makes sense to also update this in our implementation!

All the files changed are just `.prot` source files for Turnt tests, except for an updated Pest parser grammar which now mentions `prot` instead of `fn`. 

Addresses #128 (partially).


Nikil-Shyamsunder:
I'm happy with this, unless you want to consider making it `prot fn`? Then if we ever introduce compositional calls we can just list those as `fn` in the grammar without having to modify all the tests again like we have to do in this PR. it's lows stakes either way, so doesn't matter too much, but thought I would ask. Thanks!

Issue #179, opened 2026-02-12
[Monitor] do not print out thread IDs by default


ekiwi:
It would make it easier to change the monitor scheduler if we did not print out thread IDs, especially in the context of test output,

PR #181, opened 2026-02-14
[Monitor] Chore: Suppress thread IDs in output by default + use the same syntax as `.tx` files when printing traces 

ngernest:
This PR addresses #179 by adding a new CLI flag for the monitor to enable printing thread IDs for each protocol in the monitor's output (by default, thread IDs are suppressed). This makes us easier to see if future changes to the monitor's internal scheduler affect its output.

This PR also adds semicolons to the end of each transaction int he monitor output so that the syntax is exactly the same as the user-supplied `.tx` files for the interpreter -- this makes it easier to perform interpreter-monitor round-trip tests (#180).


Nikil-Shyamsunder (commented):
I updated this to reflect the syntax on #186. if you think the syntax could be improved, we can sync on slack or go ahead and make the change across both PRs.

PR #187, opened 2026-02-15
Round-trip Testing

Nikil-Shyamsunder:
This PR focuses on infrastructure, not semantic bug fixes.

We now integrate roundtrip checks with Turnt instead of running only a standalone script. 

- Added a Turnt `roundtrip` environment for roundtrip
- Roundtrip command runs per-`.tx` file via: `scripts/roundtrip_case.py {filename}`
- Added a convenience entry in `justfile` to run roundtrip through Turnt.
- CI now runs roundtrip tests

For each `.tx` file:

1. Parse `// ARGS` and `// RETURN` metadata.
2. Skip tests with non-zero `// RETURN` (recorded as skip).
3. For each trace in the .tx file, run interpreter to generate waveform
4. Run monitor on generated waveform using the same protocol context.
5. Compare expected trace block from `.tx` against monitor-emitted trace block(s) after normalization (get rid of comments, whitespace etc).
6. Emit success/skipped/failure messages

Turnt snapshots roundtrip stdout into `.rt` files 

Updated the turnt output to also have the expected (monitor) traces and the actual (interpreter) trace. so for a succeeding test, it might look like:

```
trace_block: 0
trace_result: PASS
matched_monitor_trace_index: 0
interpreter_trace:
trace {
    multiple_assign(1, 1);
    multiple_assign(1, 1);
}

parsed_monitor_trace_candidates:
candidate_monitor_trace: 0
trace {
    multiple_assign(1, 1);
    multiple_assign(1, 1);
}

candidate_monitor_trace: 1
trace {
    multiple_assign(1, 1);
    two_fork_err(1, 1);
}
```
This is good for passing examples, because we can check if there is any change in interpreter or monitor behavior that changes how many or what traces show up even for passing tests, instead of a blanket "pass".

A roundtrip trace might fail like this due to a monitor error:
```
trace_block: 0
trace_result: FAIL
failure_kind: monitor_error
interpreter_trace:
trace {
    add(1, 2, 3);
    add(4, 5, 9);
}

monitor_error:
All schedulers failed: No transactions match the waveform for DUT `Adder`
Failure at cycle 1: No transactions match the waveform in `.roundtrip_tmp/adders-adder_d1-both_threads_pass_0.fst`.
Possible transactions: [add, add_fork_early, add_doesnt_end_in_step, add_incorrect, add_incorrect_implicit, wait_and_add]
Error: Monitor failed
```

Or it might fail where the actual trace is not in the monitor output, like this:
```
trace_block: 0
trace_result: FAIL
failure_kind: trace_mismatch
interpreter_trace:
trace {
    reset();
    push(2);
    pop(2);
    idle();
    push(3);
    pop(3);
}

parsed_monitor_trace_candidates:
candidate_monitor_trace: 0
trace {
    reset();
    push(2);
    pop(2);
}
```

monitor failures due to panics are sanitized to get rid of machine-specific thread numbers (observed this on the GitHub actions CI machine) and because the line numbers in the `panicked at` line will change constantly.

we might also consider putting in a low-priority issue that some of this can be ported into rust (I am thinking specifically of the logic for generating a trace, passing it to the monitor, comparing the resulting traces, etc.) so that we have native support instead of relying on janky string processing.

### DIFF
--- a/protocols/tests/adders/adder_d1/busy_wait_fail.rt
+++ b/protocols/tests/adders/adder_d1/busy_wait_fail.rt
@@ -1,1 +1,1 @@
-SKIP: non-zero // RETURN
+SKIP: allowed to fail round trip


ngernest:
Looks good to me, thanks for working on this! I just updated the `turnt.toml` file to use `uv run` instead of `python3` to invoke the round-tripping script (to maintain consistency with the other Python scripts in the repo). 

I wonder if we should try to resolve all the failing round-trip tests in this PR or save them for another PR -- thoughts?

ngernest (approved):

scripts/roundtrip_case.py
ngernest: Do you know how hard it would be to generate a diff of the original user-supplied trace / the monitor's inferred trace if the round-trip fails? I think this would be helpful for debugging round-trip test failures. (If it's very hard, no worries! We can figure something out)

scripts/roundtrip_case.py
```diff
@@ -0,0 +1,239 @@
...
+            normalized_args.append(str(int(arg[2:], 16)))
+        elif re.fullmatch(r"\d+", arg):
+            normalized_args.append(str(int(arg, 10)))
+        else:
+            normalized_args.append(arg)
+
+    return f"{fn_name}({', '.join(normalized_args)});"
+
+
+def parse_trace_blocks(text: str) -> list[list[str]]:
```
ngernest: If you could add doc-strings for these Python strings, this would be great, thanks! (Just in case we need to modify this script in the future)

scripts/roundtrip_case.py
Nikil-Shyamsunder: It is easy to do, but I'm not sure what the best manner is to integrate that with turnt? from what i understand, turnt only produces an output file if the expected return code is provided by the process. but, in this case the expected return code is 0 and the failing roundtrip tests are returning nonzero error codes.

We could just have the script always return 0 and only have the output comparison be purely what determines if a test is passing or failing. Is there a better way than this?

Or we could keep what we have no, and one has to manually check the output by running `uv run roundtrip_case` to get the actual trace diff. What do you think is better?

Nikil-Shyamsunder: yeah that should be pretty quick change; i actually have it in my stash somewhere lol. thanks for the feedback! will request review once i make those changes

Nikil-Shyamsunder:
> Gotcha, I think Kevin @ekiwi normally prefers for all CI checks to be passing before merging a PR, I'll defer to him re: whether we can leave some round-trip tests failing when we merge this PR (i.e. fix them in a separate future PR).

Agreed. the check will "pass" now (once i fix one more bug) but it will be passing with some examples having `trace_result: FAIL` that we can either resolve before merging this or resolve later. Agree to defer to kevin. Thanks so much for the great feedback on this review

protocols/tests/adders/adder_d1/wait_and_add_correct.rt
ngernest: TODO for Ernest: check why monitor is failing here

protocols/tests/adders/adder_d0/add_combinational.rt
ngernest: TODO for Ernest: investigate this (`b` ought to be defined)

PR #199, opened 2026-02-23
(WIP) Add Tree-Sitter grammar

ngernest:
(This is purely for fun) I am trying to migrate away from VS Code and instead use the Zed editor, which only supports Tree-Sitter for syntax highlighting as opposed to VS Code's TextMate format, so this allows for syntax highlighting in Zed.


ngernest:
Closing this PR b/c it is more work than it's worth

PR #236, opened 2026-06-01
chore: Move symbol table logic to separate file

Nikil-Shyamsunder:
- the choice of AST/IR and choice of symbol table are separate things 
- the AST logic and symbol table logic are in one file right now `ast.rs` and can easily be separated. This PR does that
- the import resolution changes across the codebase were done using `cargo fix` which takes `rustc`s suggestions for the imports. I used a special nightly feature with the options `imports_granularity = "Module"` `imports_layout = "Horizontal"` to maintain our current import style


ekiwi (approved):
Seems fine

PR #237, opened 2026-06-02
IR and Straight-line code lowering 

Nikil-Shyamsunder:
Implements an IR that reflects #235. A few things on the IR I want to point out:
- Currently, nodes and ops have ids, but ops may not need them since they are quite simple, non-recursive, objects. The reason for ops, actions, and transactions to have ids is so that we can have secondary maps pointing to debug information. All 3 of these types might reasonably have debug info (though I haven't implemented any of the propogation logic yet and it becomes complicated when we start collapsing edges. If we don't give them IDs, I am not sure what the best method to maintain the debug info is.
- There is a lot of shared metadata (protocol name, args, dont_care_id, etc.) that is shared between the AST and the IR. Copying these to the new IR also requires us to add some accessors for the AST fields that probably shouldn't be exposed to client. I wonder if we should have a separate metadata structure since this stuff doesn't change as we lower the IR.
- I added a special `true_id` to the private state of the IR, so that we can easily construct trivial guards. Maybe there is a better pattern for doing this.
- Now, we have `ast::Protocol` and `ir::Protocol`. In `lowering.rs`, I wrote the imports so that each structure is referred to with `ast::` or `ir::` in order to make things abundantly clear. I'm wondering if I should be naming these two structs differently instead.

There is also a simple lowering pass implemented:
- it handles only straight line code
- I figure it is best to make this construction as trivial as possible, and deal with collapsing edges later. Specifically,
    - it maintains an empty entry node and an exit node that only has the `Done` action
    - `step()` calls create their own nodes with no actions and a trivial step transition
    - There are tests of the lowering: 2 using inline protocols, 1 using the `add_d1.prot` file. I put these in `ir.rs` because they are directly inspecting the IR. I feel they should go in `lowering.rs`, but I think it will add some unecessary accessors. I would like to know if there is a good pattern for this in Rust.


protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Default)]
+// ops (so far) are non-recrusive, may not be necessary to have an ID, but useful for debug maps
+// if we want debug info on expressions and transitions, we'll need a way to do this with or without Ids
+pub struct OpId(u32);
+entity_impl!(OpId, "op");
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// an Action is a guard and an operation to perform when the guard is true
+pub struct Action(pub ExprId, pub OpId);
```
ekiwi: Instead of describing the fields in a comment, it might make sense just to name them.

So:
```rs
pub struct Action { guard: ExprId, op: OpId }
```

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// A Transition is a guard, a target node, and flag if this transition consumes step
+pub struct Transition(pub ExprId, pub NodeId, pub bool);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+// A node is a (logically) unorded set of actions and transitions
+pub struct Node(Vec<Action>, Vec<Transition>);
```
ekiwi: For this one, I definitely want you to name the fields: 
```rs
struct Node { actions: Vec<Action>, transitions: Vec<Transition> }
```

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,428 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// A Transition is a guard, a target node, and flag if this transition consumes step
+pub struct Transition {
+    guard: ExprId,
+    target: NodeId,
+    consumes_step: bool,
+}
+
+impl Transition {
```
ekiwi: Are you sure, you need all these "wrapper" methods? Is there much that you gain over just making the fields public?

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,428 @@
...
+
+impl Node {
+    pub fn empty() -> Self {
+        Self {
+            actions: vec![],
+            transitions: vec![],
+        }
+    }
+
+    pub fn actions(&self) -> &Vec<Action> {
```
ekiwi: Is this getter needed?

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,428 @@
...
+        &self.actions
+    }
+
+    pub fn transitions(&self) -> &Vec<Transition> {
```
ekiwi: same

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,428 @@
...
+    AssertEq(ExprId, ExprId),
+    Fork,
+    Done,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+// TODO: A lot of this is metadata shared between the AST and IR.
+// I wonder if it is worth factoring out a metadata between the AST and IR.
+// right now I'm distinguishing ast::Protocol and ir::Protocol but i dont know if there is a better naming convention.
+pub struct Protocol {
```
ekiwi: You could call this struct `ProtoGraph` in order to distinguish it from the AST-level `Protocol`

protocols/src/ir/ir.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub entry: NodeId,
+
+    /// Optional type parameter (identified by its `SymbolId`)
+    pub type_param: Option<SymbolId>,
+
+    /// Whether the protocol has been marked as `idle` with `#[idle]`
+    pub is_idle: bool,
+
+    /// Maps `ExprId`s to their corresponding `Expr`s
+    exprs: PrimaryMap<ExprId, Expr>,
```
ekiwi: It might make sense to refactor all the expression handling into its own struct that can be shared between AST and IR.

protocols/src/ir/ir.rs
Nikil-Shyamsunder: I switched the fields to public.

protocols/src/ir/ir.rs
Nikil-Shyamsunder: these are needed for the graphviz (iterates over the actions). I can also pass an iterator instead of actually passing the vector.

protocols/src/ir/ir.rs
Nikil-Shyamsunder: see above

protocols/src/ir/proto_graph.rs
```diff
@@ -0,0 +1,415 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Op {
+    Assign(SymbolId, ExprId),
+    AssertEq(ExprId, ExprId),
+    Fork,
+    Done,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+// TODO: A lot of this is metadata shared between the AST and IR.
```
ekiwi: is this TODO now done? With the introduction of the `ProtocolContext`?

Nikil-Shyamsunder: yup, thanks for pointing out. I removed the comment.

ekiwi (approved):
At this point, I think you should just get this merged + also merge the Graphviz backend.

Next up would be:
- hook this up into the `cli` tool so that you can run tests with `turnt`
- write a simple interpreter so that we can think more about the semantics and test them

PR #238, opened 2026-06-02
Serialize Automata IR to DOT file for graphviz

Nikil-Shyamsunder:
- Wrote a very simple serializer for the IR. Right now, the file is called `graphviz.rs` but I don't know if I should change it to also be called `serialize.rs` (like in the frontend), or maybe `serialize_dot.rs` instead
- Expressions are serialized using the existing expression serializer Francis wrote way back when. Ops have a custom, but very simple serialization
- The DOT visualization is left to right with a clearly labelled start node, actions within each node (if there are any), and transitions as edges between nodes, labelled with their guard and `/ step` if they are a step-edge. Note that `[1]` denotes a guard of `1` (true). not sure if we should have a special serialization for `1/0` as `T/F`.
- Tested with a checked-in snapshot of the adder protocol, producing the following viz:

image
- Since this new IR is quite simple, I expect this serialization code to be robust to when we start doing graph rewrites and tackle control flow as well. I foresee we may need to make some tweaks for formatting (maybe add colors, etc.) when graphs start to get larger

### DIFF
--- a/protocols/src/frontend/serialize.rs
+++ b/protocols/src/frontend/serialize.rs
@@ -5,6 +5,7 @@
 // author: Francis Pham <fdp25@cornell.edu>
 
 use std::io::Write;
+use std::ops::Index;
 
 use baa::{BitVecOps, BitVecValue};
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


ekiwi (approved):
Looks good!

PR #240, opened 2026-06-04
Graph Interpreter 

Nikil-Shyamsunder:
Adds a simple graph interpreter for the ProtoGraph IR. 

**Semantics:**
- start at the entry node. there is a sticky assignment buffer that starts with all input pins as `DontCare`.
- first, go through the actions that correspond to assignments and update the buffer. Then reapply all pins with the updated values. Then evaluate all other actions. 
- then evaluate the transitions. 
- Note that all of these ordering decision (update assignments --> all other actions -->  transitions) stop mattering if we don't have combinational dependency issues, which I don't believe we aren't worrying about right now. 
- For now, there is a check that the done flag is raised if and only if there are no triggered transitions. This, i expect to stop working when we introduce repeat/for _ in loops, but as a sanity check I am keeping it for now. There is also a check that there are not multiple transitions triggered (non-determinism) - once again, this may be violated in the future but we will cross that bridge when we get there.
- The way of printing results is much simpler than the AST interpreter (just using assertions and panics for now, stripping the thread info that is non-deterministic in order for snapshots to stay the same over time). 

**Practical matters on engineering the graph interpreter:**
I went about this in a way in order to minimize code duplication between the AST interpreter and IR interpreter (by taking advantage of the common Patronus setup, simulator interface functions, and expression evaluation). The strategy is as follows: 
- In `ast.rs`, allow for the creation of an empty AST given a `ProtocolContext` wth a `from_context` constructor
- In `interpreter.rs` (the AST interpreter), make two functions public: apply simulator inputs and evaluate expressions. 
- In `graph_interpreter.rs`, create a an empty AST using the `from_context` constructor, then use those two public functions to evaluate expressions and drive inputs for the sake of the graph interpreter semantics, ignoring the rest of the capabilities of the AST interpreter.

The change surface ends up being quite small, which is nice, though there is some janky-ness involved I admit. If we feel this is two janky right now or wish to improve this in the future, I can address it now or we can file an issue respectively. I expect the best way to do this is to split the AST interpreter into the interpreter and a shim, and reuse that shim across the AST and graph interpreters. It's not clear to me how much extra work this will be. Additionally, I wish to avoid doing too much here if we end up switching to Patronus expressions.

**Testing:**
- Added CLI command to lower from `.prot` files all the way to the IR and serialize DOT files
- Added CLI command to interpreter from `.tx` at the graph IR level. 
- Set up a turnt environment for both of these. Because most tests will fail with turnt for now, I set up an `allowlist` which `just graph_interp` reads and uses to launch the allowlisted tests. This isn't in CI yet (though I can turn it on if we want). Also, only one test is allowlisted for now. I will start trying more tests, but I wanted to get this PR up first. 
- For now, I do not think we should use turnt for the serialization tests (it's going to be one extra file in the test directory for every `.prot` + a second allowlist for .prot fles), and serialization tests have historically been done with insta which I think is fine, and insta allows us to have these in a separate directory. The add_d1.prot file is already in the insta snapshot tests. Once again, will start adding more tests to the turnt allowlist/insta tests as we go.

### DIFF
--- a/cli/src/bin/graph-interp.rs
+++ b/cli/src/bin/graph-interp.rs
@@ -1,0 +1,122 @@
+use std::panic::{AssertUnwindSafe, catch_unwind};
+
+use clap::Parser;
+use protocols::frontend::ast::Protocol;
+use protocols::frontend::diagnostic::DiagnosticHandler;
+use protocols::frontend::symbol::SymbolTable;
+use protocols::ir::graph_interpreter;
+use protocols::ir::lowering::lower_ast_to_ir;
+use protocols::setup::create_sim_context;
+use protocols::{Value, frontend, transaction_frontend};
+use rustc_hash::FxHashMap;
+
+#[derive(Parser, Debug)]
+struct Cli {
+    /// Paths to one or more Verilog (.v) files
+    #[arg(long, value_name = "VERILOG_FILES", value_delimiter = ' ', num_args = 1..)]
+    verilog: Vec<String>,
+
+    /// Path to a Protocol (.prot) file
+    #[arg(short, long, value_name = "PROTOCOLS_FILE")]
+    protocol: String,
+
+    /// Path to a Transactions (.tx) file
+    #[arg(short, long, value_name = "TRANSACTIONS_FILE")]
+    transactions: String,
+
+    /// Name of the top-level module (if one exists)
+    #[arg(short, long, value_name = "MODULE_NAME")]
+    module: Option<String>,
+
+    /// Skips the static checks for step/fork errors.
+    #[arg(long)]
+    skip_static_step_fork_checks: bool,
+}
+
+fn load_protocols(cli: &Cli) -> Vec<(Protocol, SymbolTable)> {
+    let mut handler = DiagnosticHandler::new(clap::ColorChoice::Never, false, true, false);
+    frontend(
+        &cli.protocol,
+        &mut handler,
+        cli.skip_static_step_fork_checks,
+    )
+    .unwrap()
+}
+
+fn load_traces(cli: &Cli, protos: &[(Protocol, SymbolTable)]) -> Vec<Vec<(String, Vec<Value>)>> {
+    let mut handler = DiagnosticHandler::new(clap::ColorChoice::Never, false, true, false);
+    transaction_frontend(&cli.transactions, protos.iter(), &mut handler).unwrap()
+}
+
+fn build_arg_map<'a>(
+    args: &'a [protocols::frontend::symbol::Arg],
+    symbols: &'a SymbolTable,
+    values: Vec<Value>,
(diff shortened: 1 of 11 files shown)


cli/src/main.rs
```diff
@@ -2,26 +2,49 @@
...
+use protocols::ir::lowering::lower_ast_to_ir;
+use protocols::setup::create_sim_context;
 use protocols::{Value, frontend, transaction_frontend};
+use rustc_hash::FxHashMap;
 
 #[derive(Parser, Debug)]
 struct Args {
     #[arg(short, long, value_name = "PROTOCOLS_FILE")]
-    protocol: String,
+    protocol: Option<String>,
```
ekiwi: The `cli` is build around the assumption that every command operates on at least one protocol file. If your command does not require a protocol file, then it might not be appropriate to add it to the `cli`.

Can you elaborate on why you had to make this argument optional?

cli/src/main.rs
```diff
@@ -164,23 +187,167 @@ fn run_verilog_tb(
...
 
-    // we always parse and type check the protocol file
-    let skip_static_step_fork_checks = false;
-    let mut d = DiagnosticHandler::new(ColorChoice::Auto, false, true, false);
-    let protos = frontend(args.protocol, &mut d, skip_static_step_fork_checks).unwrap();
-
     match args.command {
-        None => {}
+        None => {
+            let protocol_filename = args.protocol.expect("missing --protocol");
```
ekiwi: We are getting a lot of redundance here because you made the `args.protocol` optional and handle it separately for each command. Why did you have to do that?

ekiwi (changes requested):
Let's see if we can reduce the number of changes to the `cli`. Everything else looks good.

cli/src/main.rs
Nikil-Shyamsunder: yup! this is the idea:
-  We wanted a new Turnt environment for graph interpretation that behaves like the existing `interp` environment
- we did not want a separate binary; instead we wanted to use the `cli` binary and add a new subcommand
- the interp command looks like: `cargo interp --transactions {filename} {args}`, where the args are picked up by `turnt` from the `.tx` file 
- for the graph interpreter, since we want to reuse the same args, I needed the same model, something like: `cargo run protocols-cli -- run-graph {filename} {args}`
- the problem is that the --protocol flag is expected at the top-level, but in order to allow us to handle it at the subcommand level so we can stay in line with the ast `interp` args format, the `--protocol` needs to be set at the subcommand level, so it needed to be switched to optional. 

I don't think there is an easy fix in Turnt. Instead, I am looking at the clap docs right now and our use case looks to match this `global` arg thing: https://docs.rs/clap/latest/clap/struct.Arg.html#method.global:~:text=Assume. I will try this

ekiwi (approved):

PR #244, opened 2026-06-15
Edge contraction

Nikil-Shyamsunder:
I was playing around with how to do the edge contraction, came up with something very simple, and it seems to be doing what we want for now better than I initially expected. This is mainly vibes based as a starting point to play around with. Here is the general idea

**`edge_contraction.rs`**
- let's decompose the Action-specific combination stuff (e.g. how do we combine two actions together, how do we combine two forks together) from the edge contraction
- if you ignore said action stuff, edge contraction is extremely simple. do the following until you can't anymore:
   1. Take an edge that is not a step edge
   2. AND all the actions in the to-node with the transition guard from the from-node and stick them into the from-node
   3. AND all the transitions *from* the to-node with the transition guard from the from-node and stick them into the from-node 
   4. remove the transition from the from-node to the to-node
- Once we do this, all the smart action/transition combination stuff can be done as intra-node reasoning/reasoning about just the out-edges of a node. If we merge in a way that preserves the program order from the AST, then we can reason within a node about actions post-contraction in order. It's possible this will end up being the wrong approach, but I wanted to get something done to experiment with. we can also start thinking about adding internal state to the automata - the first piece of internal state might be a "previous assignment" local var for every input, so that we can assign that value if there is no overridden action. 
- for some reason I have an intuition doing this from right-to-left (approximately reverse node insertion order and reverse transition insertion order) minimizes duplicate edges. I will have to experiment with this more. This is purelt vibes based.

**graphviz.rs**
- there is no garbage collection of unused nodes when we do edge_contraction
- this is OK for now, but it's very annoying to look at graph visualizations with a bunch of unreachable nodes in the digraph. Previously, when generating the graphviz we just iterated over all the nodes. Now, we just do a BFS from the entry so we only serialize reachable nodes.

**repo refactor for `And`:**
-- added support for `And` as a binary operation in order to produce graphs using it. some places I filled it in with implementation details in the match arm (when I knew very easily how to implement it), other places I just left it `todo!(...)`

**tests:**
- I manually checked the serialized test cases and they look correct pre- and post-contraction. For brevity, here are two examples

**simple straight-line adder:**
image

**counter:**
**I am surprised this worked correctly.** What's interesting here is we have an assertion that can get executed right after the while *without a step in between*. If the while never trips, the `step()` never runs. So, to handle the assertion triggering without a step if the while runs at least once or not is a bit weird. But, as you can see from the graph, the assertion propogates correctly to make this work. 

```
prot count_up_extra_assertion<dut: Counter>(a: u64) {
    // Loading dut with values
    dut.a := a;

    while !(dut.s == dut.a) {
        step(1);
    }

    // something after the loop with no step() in between
    assert_eq(dut.s, a);

    // now, dut.s == dut.a. Since s >= a, stepping once
    // more should result with dut.0 equal 0
    step(1);
    assert_eq(dut.s, 64'b0);

    fork();
    step();
}
```

image

**Next steps:**
- add an example with a loop that doesn't have a step 
- these contracted IRs have only gotten an eye-test from me. we should differentially test them with the pre-contraction IRs using the graph interpreter
- think about action/transition combination rules
- think about the holding assignment semantics (if we want to now).


protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Contract non-step edges from right to left until no more eligible edges remain.
+pub fn contract_edges(protocol: &mut ProtoGraph) {
+    let node_ids = protocol
+        .nodes()
+        .map(|(node_id, _)| node_id)
+        .collect::<Vec<_>>();
+
+    // iterate nodes
+    for lhs_id in node_ids.into_iter().rev() {
```
ekiwi: To me, it would be a lot clearer if we could refer to the nodes either as destination (dest) / target or prev(ious) / next instead of lhs/rhs.

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,70 @@
...
+            // create actions with the transition to RHS guard & the action in RHS guards
+            let merged_actions = rhs_actions
+                .into_iter()
+                .map(|action| Action::new(and_guard(protocol, lhs_guard, action.guard), action.op))
+                .collect::<Vec<_>>();
+            // create transitions with the transition to RHS guard & the transition from RHS guards
+            let merged_transitions = rhs_transitions
+                .into_iter()
+                .map(|rhs_transition| {
+                    Transition::new(
```
ekiwi: It might make sense to add a helper method to `Transition`. Something like `withAdditionalGuard(self) -> Self`.

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,70 @@
...
+        .collect::<Vec<_>>();
+
+    // iterate nodes
+    for lhs_id in node_ids.into_iter().rev() {
+        let mut transition_idx = protocol[lhs_id].transitions.len();
+
+        // iterate transitions backwards. for some reason I feel this will prevent
+        // duplicate edges; revisit when we have evidence of things going wrong
+        // TODO: the backwards iteration logic here is a bit messy
+        while transition_idx > 0 {
```
ekiwi: In an attempt to simplify the iteration logic.
- Let's create two vecs for transitions, one for transitions that have a step and one for "instant" transitions.
- Now the loop can  be: `while let Some(transition) = instant_transitions.pop() {`
- When you add new transitions, you will have to decide which `Vec` to put them into based on whether they include a step or not

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,70 @@
...
+                        rhs_transition.target,
+                        rhs_transition.consumes_step,
+                    )
+                })
+                .collect::<Vec<_>>();
+
+            // modify the LHS nodes with these
+            let lhs_node = protocol.node_mut(lhs_id);
+            lhs_node.transitions.remove(transition_idx);
+            lhs_node.actions.extend(merged_actions);
```
ekiwi: You will have to add some code here to check each new action against the existing actions to see if there is an ordering that you need to maintain. I.e., when the actions are in two separate nodes, there is a clear order in which they are executed. But the actions in a single node are treated as _unordered_. Thus, you have to merge actions (or fail an assertion for now) if there are two actions that might lead to different behavior depending on the order in which they are executed.

ekiwi (changes requested):
This looks like a good first step. Take a look at me comments.

protocols/src/ir/edge_contract.rs
Nikil-Shyamsunder: will address in follow-up

protocols/src/ir/edge_contract.rs
Nikil-Shyamsunder: done

protocols/src/ir/edge_contract.rs
Nikil-Shyamsunder: done

protocols/src/ir/edge_contract.rs
Nikil-Shyamsunder: done

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+pub fn contract_edges(protocol: &mut ProtoGraph) {
+    let node_ids = protocol
+        .nodes()
+        .map(|(node_id, _)| node_id)
+        .collect::<Vec<_>>();
+
+    for source_node_id in node_ids.into_iter().rev() {
+        // all transitions out of source begin as pending
+        let mut contracted_actions = protocol[source_node_id].actions.clone();
+        let mut step_transitions = Vec::new();
```
ekiwi: Can you already add the `step_transitions` from the original node here?

You can use: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+        let mut pending_transitions = protocol[source_node_id].transitions.clone();
+
+        while let Some(transition) = pending_transitions.pop() {
+            // if a step transition, don't do anything with it. just save it
```
ekiwi: Once you do the change above, remove the following code and replace it with a `assert!(!transition.consumes_step);`

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+
+            // get the info on the target node
+            let target_node_id = transition.target;
+            let incoming_guard = transition.guard;
+            let target_actions = protocol[target_node_id].actions.clone();
+            let target_transitions = protocol[target_node_id].transitions.clone();
+
+            // merge actions from the target node into the source node
+            // with the source transition guard AND-ed on
+            let contracted_target_actions = target_actions
```
ekiwi: I am often a fan of using `map`, but here, things would be clearer with a good old for loop:
```rs
for action in target_actions {
  contracted_actions.push(Action::with_guard(&action, and_guard(protocol, incoming_guard, action.guard)));
}
```

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+                .into_iter()
+                .map(|action| {
+                    Action::with_guard(&action, and_guard(protocol, incoming_guard, action.guard))
+                })
+                .collect::<Vec<_>>();
+            contracted_actions.extend(contracted_target_actions);
+
+            // merge transitions out of the target node into the source node
```
ekiwi: Same here: instead of a `map` followed by a `for`, just use a plain `for` loop.

Something like:

```rs
for target_transition in target_transitions {
  let contracted_transition = Transition::with_guard(
                        &target_transition,
                        and_guard(protocol, incoming_guard, target_transition.guard),
                    );
  if new.consumes_step {
    step_transitions.push(contracted_transition);
  } else {
   pending_transitions.push(contracted_transition);
  }
}
```

protocols/src/ir/edge_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+            let target_actions = protocol[target_node_id].actions.clone();
+            let target_transitions = protocol[target_node_id].transitions.clone();
+
+            // merge actions from the target node into the source node
+            // with the source transition guard AND-ed on
+            let contracted_target_actions = target_actions
+                .into_iter()
+                .map(|action| {
+                    Action::with_guard(&action, and_guard(protocol, incoming_guard, action.guard))
+                })
```
ekiwi: I see that you aren't checking actions for interference yet. Are you planning to do that in the next PR?

ekiwi (changes requested):

protocols/src/ir/edge_contract.rs
Nikil-Shyamsunder: yes

PR #259, opened 2026-06-25
Differential testing for graph interpreter

Nikil-Shyamsunder:
- Edge contraction is now done under the flag `--contract-edges` in the graph interpreter CLI
- Pre and post edge contraction cases (i.e. run the interpreter with and without the flag) are generated in `generate_rust_configs.py` using the same expect file, so if they have different  behavior runt will give an error
- `fifo.tx` relies on fork support. some of the transactions succeed, but because they `DontCare` certain pins expecting a follow-up transaction will give proper values, they succeed but leave the DUT in a "corrupted" state for following transactions. These tests are simply filtered out right now since we're not worrying about fork (just yet)
- A few tests ended up failing after edge contraction because there was a bug in the graph interpreter. At the beginning of a new node, we weren't reading the latest values from the simulator; this is a problem at step boundaries leading to guards and actions being evaluated on stale DUT values. 
- In the AST interpreter, we were not correctly handling the case of "If a DUT assigns a value to an input port, and we then observe a combinationally dependent output *or that input itself*, we cannot reassign the port." We were handling combinational outputs, but not including the input pin itself. We were special casing and differentiating inputs/outputs in a few places where it just made sense to handle it universally by treating inputs as combinationally dependent on themselves and treating observation identically for both SymbolKinds. I fixed this bug and modified the test that should've been failing on this.
- A few `.graphout` files were still around (from when we used turnt) - deleted them


protocols/src/tests/snapshots/protocols__frontend__typecheck__tests__add_d1.snap
```diff
@@ -21,13 +21,13 @@ warning: Inferred RHS type as u32 from LHS type u32.
    │   ^^^^^^^^^^^ Inferred RHS type as u32 from LHS type u32.
 
 warning: Inferred RHS type as u32 from LHS type u32.
-   ┌─ tests/adders/adder_d1/add_d1.prot:83:3
```
ekiwi, line 24: This isn't actionable for this PR, but maybe, we should disable this type inference warning. It only creates noise.

ekiwi (approved):
Looks good.

PR #262, opened 2026-06-25
Improve type names with latest terminology

Nikil-Shyamsunder:
Our terminology has changed and standardized a bit since the old days as we have been writing and talking about protocols more. The AST interpreter in particular was way behind terminology wise, and because I need to plug in to some of those types in order to build out the NFA to DFA thing with concrete transaction info, I thought I'd take a sec to fix this since it's been bothering me.

- A Protocol is a Protocol (this used to be called IR and was fixed at the type level but not in the variables)
- A ProtocolInfo is a triple of the form (Protocol, SymbolTable, NextStmtMap) - this was called TransactionInfo
- A Transaction is an instance of a protocol
- An Invocation is a (String, Vec<Values>) representing an call to run a protocol with concrete args. One invokes a protocol with concrete args to instantiate a transaction. - this used to be called a Todo
- a few more little things like variable names corresponding to values of these types
- a few places we either had (1) comments that were self-evident from the code (2) missing comments for some important things. I didn't fix all of these but at least the ones I was already looking at.

### DIFF
--- a/interp/src/main.rs
+++ b/interp/src/main.rs
@@ -155,7 +155,7 @@ fn main() -> anyhow::Result<()> {
             &mut protocols_handler,
             cli.max_steps.unwrap_or(u32::MAX),
         );
-        let results = scheduler.execute_todos();
+        let results = scheduler.execute_transactions();
         let trace_failed = results.iter().any(|res| res.is_err());
 
         if trace_failed {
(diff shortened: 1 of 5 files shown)


ekiwi (approved):
Looks good. I am not sure whether there actually is a meaningful distinction between a "protocol invocation" and a "transaction". But we can always harmonize that later.

PR #264, opened 2026-06-27
Full Trace Lowering

Nikil-Shyamsunder:
Section 5 of this document (https://github.com/cucapra/automata-paper/blob/main/main.pdf) details how this algorithm works.

- Sections 4.3 and 5.2 detail the node merging semantics. All of this logic (for both) is in `edge_contract.rs`, which distinguishes between ordered and unordered merges via a Boolean passed in.
- `lowering.rs` is now stateful, allowing you to lower multiple ASTs into a shared IR. You can call `self.lower_fragment(ast)`, and it mutates the IR by adding the new protocol, though this new protocol isn't connected at all to any existing nodes in the IR. it the returns to you the entrance, exit, vector of nodes for the lowered fragment which you can modify and wire up as you please; for the standard symbolic lowering, that's just setting the IR entry to the lowered fragment's entry. for the concrete trace lowering, it's merging the new transaction in at the graft points of the previous transaction.

- **Full trace lowering stuff** is in `trace_lowering.rs`. The algorithm is pretty simple here. There is a set of graft points which is just the fork and done nodes for every transaction we lower. We lower each AST into nodes, subbing in concrete values and performing the edge contraction algorithm on it. Note that technically we call edge contraction on the *entire* IR, but since everything we've already processed has no nonstep edges, it is idempotent to call it again on those. This is a bit wasteful, so I put a TODO down, but its not a logical issue for now. We then "graft" it into whatever the existing automata is by merging the entry node it into the graft points of the previous transaction ( copy of the lowered transaction at each graft point). Then we're done building an NFA
    - A problem with this lowering system is that the `ProtocolContext` of a `ProtoGraph` was initially a bunch of stuff about the AST it was lowered from (name, type param, args etc.). This is not really meaningful for a full concrete trace or a symbolic multi-protocol graph, so we may need to change this at some point as we have more general applications that use a single ProtoGraph. I added a TODO for this. 

- **NFA to DFA:** The subset construction algorithm is in `determinize.rs`. The algorithm itself is pretty straightforward. Most of the difficulty is back in `edge_contract.rs` where we have the unordered merge semantics. I would take a look at `merge_unordered_assignment_rhs` - this is the core change. One thing we discussed today is right now our algorithm builds a new `PrimaryMap` of nodes, and then subs them into the IR. This is not exactly true; actually it clones the graph using a `from` function where we can pass a new set of nodes. This isn't perfect and we can do instead what we discussed today of just leaving the old nodes around, but I have a stack of commits above this PR that I don't want to mess up too badly (will address in the follow-up). 

- **Assignment merging:** Basically, we have to break down the ITE, and then move the concrete assignments up front (since across nodes concretes should always be preferred against DontCares, but within a node we should respect the user ordering). The exact algorithm is in the code but also detailed in 5.3; it is a little unfortunate we have to break down the ITEs and then reconstruct them like this, but we've committed to the ITE / always one assignment of the port per node thing so I'm gonna keep it this way for now. All the breakdown and build-back-up logic is in one function call, so an IR is never produced that breaks this invariant. This logic took me a while to get right, but I think it is correct and I decided to make unit tests specifically for this logic because it is important and a bit confusing + hard to test  these edge cases with actual protocols. 

Example of the merge semantics:
Suppose node A has `[1] DUT.a := ite(p, X, ite(q, 1, DUT.a))`
Suppose node B has `[r] DUT.a := 2`

When merging, we want to prefer B's concrete assignment to A's DontCare, but we want to prefer A's DontCare to A's Concrete due to intra-transaction ordering. The way we can do this algorithmically is to break down the ITE into:

```
[p] DUT.a := X
[!p /\ q] DUT.a := 1 
[r] DUT.a := 2
```

Then we move all the concrete assignments up front, so we get `[1] DUT.a := ite(p, X, ite(not p and q, 1, ite(r, 2, DUT.a)))`  - not pretty, but pretty sure this works. We also add an `InternalAssertFalse` for `(!p and q) and r)`.


ekiwi (approved):
Let;s get this merged. We should meet to discuss the whole pipeline next week. I am a bit lost.

PR #280, opened 2026-07-10
Diff Testing for Failing Tests

Nikil-Shyamsunder:
~500 lines of this PR is just the new runt config + the test expect files. Other than that it's pretty straight forward.

- Instead of panicking on errors, it returns a `GraphInterpreterResult` which is a waveform, a `GraphInterpreterError`, and a custom string message. A `GraphInterpreterError` can either be an `InternalAssertFailure` or `ExternalAssertFailure` (external just means an `assert_eq` fails) and in which cycle. 
- The transition system flow only supports messages of the form "{internal or external} assert failed in cycle {n}", which we can diff against by using the `--graph-brief-errors` flag so the graph interpreter gives less detailed messages (it doesnt print the custom string message from the `GraphInterpreterResult`
- I then set up a new runt configuration that is just for assertion failures and conflicting assignment errors. I put this in a separate config because the existing waveform config also compares to the AST interpreter, which we don't want. combinational dependency errors are the other big class of errors, but we aren't explicitly enforcing these right now. It diffs between just the graph interpreter and the transition system, not the AST interpreter (but I checked by hand and everything looks to be failing, and failing for approximately the right reasons).
- In doing this, I found a bug in my previous assignment propagation code in the graph interpreter. We were running one big assignment propagation at the end, but really we should be propagating for each individual IR as we graft IRs into the trace (this was my original intention but I was doing it wrong). Now the flow is for each AST we lower, we contract its edges, propagate the assignments, and then graft it on. 
- All the tests in the catalog labelled assertion or conflicting assignment failures seem to be giving the expected result!
- Added to CI.


ekiwi (approved):
Thanks. This is a good compromise.

PR #282, opened 2026-07-11
Integrate BI tests with Runt & CI

ngernest:
This PR addresses Kevin's comment [here](https://github.com/cucapra/protocols/pull/270#pullrequestreview-4645673039) in #270, adding BI Runt tests to our testing framework / CI:
> Can you make a separate PR for adding the bi to the tests and to CI? Then we can iterate on how to best integrate this with the new testing infrastrucutre. 

Most files changed in this PR are new `.expect` files containing the BI output, the key code changes to review are:
- `scripts/generate_runt_config.py` (I followed Nikil's instructions in #252 to add support for Runt tests here. Note that the BI uses the same set of test cases as the monitor, i.e. the BI test suite uses `MONITOR_CASES` in `test_catalog.py`. This is possible since BI and monitor share largely the same CLI args.)
- `bi/src/main.rs` (described below)

There are some minor changes to the `bi` executable in this PR:
- Added a new CLI arg `--color never` so that error messages are printed in plaintext (no color) in `.expect` files, like the monitor
- Following the advice here (https://users.rust-lang.org/t/suppress-panic-message/6303/3), I added a panic hook so that when `bi` panics, instead of the default Rust error message:

```rust
thread 'main' (11301631) panicked at bi/src/bi.rs:628:21:
[add_busy_wait@00!] Cannot fork at step zero!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

we instead print (and only save in `.expect` files) the following:

```rust
panicked at bi/src/bi.rs:628:21:
[add_busy_wait@00!] Cannot fork at step zero!
```

This avoids the OS thread ID (the `thread 'main' (11301631) ...` information) in the default error message from appearing in the `.expect` files, since the OS thread ID changes every time we run the executable.

**Two test infra TODOs for future PRs**
1. For the monitor (not BI), when the monitor is expected to fail (e.g. running on a Brave New World buggy waveform), the corresponding `.expect` file only displays the error code:

```
---CODE---
124
```

Ideally, we change this behavior so that monitor error messages are also captured in `.expect` files (matching what is done for `bi` in this PR). This is a one-line change in `generate_runt_configs.py`, but also affects a bunch of `.expect` files, so I wanted to defer it to a subsequent PR. 

2. Kevin mentioned in #270 that:
> You should also try to only have a single .prot for each one of the brave new world benchmarks. After all, our claim is that a single correct protocol can distinguish between buggy and fixed version. I think the reason we used to maintain two copies was because of how turnt test discovery works. With the new testing infrastructure, we should not have to do that anymore.

To implement this, we have to adapt `generate_runt_configs.py` so that for the Brave New World test cases, one single `.prot` can correspond to multiple `.expect` files (different output for buggy / fixed waveforms). This is doable, but I'd like to defer this to a future PR since it will modify (remove) a bunch of `.expect` files.


Nikil-Shyamsunder (approved):

bi/src/main.rs
```diff
@@ -86,6 +91,14 @@ struct Cli {
 
 #[allow(unused_variables)]
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // By default, a Rust panic error messages prints the OS's thread ID
+    // (e.g. `"thread 'main' (<tid>)`), but this changes every time we run this executable.
+    // To prevent the TID from appearing in error messages that appear in the expected
+    // output files for Runt, we register a custom panic hook that only prints
+    // the error message & source code location associated with the panic
+    // (via the `Display` trait for the `PanicHookInfo` type).
+    std::panic::set_hook(Box::new(|info| eprintln!("{}", info)));
```
ekiwi, line 100: We should handle this in a different way. Probably instead of panicing we would like to exit more cleanly in the error case.

scripts/generate_runt_configs.py
```diff
@@ -195,6 +195,23 @@ def monitor_runt_command(case: dict) -> list[tuple[str, str]]:
...
+def bi_runt_command(case: dict) -> list[tuple[str, str]]:
+    # We pass in `--color never` to BI to suppress color in error messages
+    # (so that the .expect files only display plaintext)
+    cmd = [*binary_prefix("bi"), "--color", "never", "--protocol", case["path"]]
+    if case["wave"]:
+        cmd += ["--wave", case["wave"]]
+    if case["instances"]:
+        cmd += ["--instances", *case["instances"]]
+    cmd += case["extra_args"]
+    if case["timeout_secs"] is not None:
```
ekiwi, line 208: Why would we ever need a timeout here?

ngernest: Some of the existing monitor test-cases in `MONITOR_CASES` (on `main`) have an expected timeout of 5 seconds, e.g. `nested_busy_wait.prot` where there are nested `repeat` loops and the monitor fails to infer both  
`outer_iters` and `inner_iters`:
```
prot nested_busy_wait<DUT: Adder>(a: u32, b: u32, outer_iters: uint, inner_iters: uint, s: u32) {
  DUT.a := a;
  DUT.b := b;
  repeat outer_iters iterations {
    repeat inner_iters iterations {
      step();
      assert_eq(s, DUT.s);
    }
    step();
    assert_eq(s, DUT.s);
  }
  DUT.a := X;
  DUT.b := X;
  assert_eq(s, DUT.s);
  fork();
  step();
}

```
Since the `bi` and `monitor` share the same test cases in `test_catalog.py`, I added this here just ot make it consistent with the monitor tests (this is also what is done in the existing [`monitor_runt_command`](https://github.com/cucapra/protocols/blob/03f184a8ac1f8278544405154c303e5914eede08/scripts/generate_runt_configs.py#L184) function in `generate_runt_configs.py`). I can get rid of this though if it is not needed for the BI -- I am thinking perhaps this is the case and this reveals a bug in the monitor.

I looked at the BI output for these test cases where the monitor times out and it says `Cannot fork at step zero!`, so I think this is somewhat related to the monitor bug #214 where it doesn't handle the case when a loop takes 0 iterations (i.e. we skip the loop) and we end up having a `fork` in cycle 0 (for the protocol above).

